PR #3, opened 2017-11-03
Looks like typo in daemonset

Bregor:
Due to 

    metadata:
      name: kured            # Must match `--ds-name`
      namespace: kube-system # Must match `--ds-namespace`

There should be 

    - --ds-name=kured
    - --ds-namespace=kube-system

As args.


awh:
Good catch @Bregor, thanks!

PR #5, opened 2017-11-17
Set variables based on correct arguments w.r.t. name of daemonset and…

SaaldjorMike:
… namespace.


awh:
Good catch, thanks @SaaldjorMike.

PR #7, opened 2017-12-04
add current version

ben-st:
add version to the DS, since there is no latest tag.

### DIFF
--- a/kured-ds.yaml
+++ b/kured-ds.yaml
@@ -11,7 +11,7 @@ spec:
     spec:
       containers:
         - name: kured
-          image: quay.io/weaveworks/kured
+          image: quay.io/weaveworks/kured:1.0.0
           imagePullPolicy: IfNotPresent
           command:
             - /usr/bin/kured


deedubs:
While this adds a maintenance cost I believe its a better alternative than promoting the use of a `latest` tag.

ben-st:
bump

ben-st:
bump, since this is now more then 2 month old.

awh:
Hi @ben-st 

I don't want to make this change for two reasons:

1. It breaks my development workflow
1. There are versioned manifests for installing releases e.g. https://github.com/weaveworks/kured/releases/download/1.0.0/kured-ds.yaml

awh:
@deedubs IMO it doesn't promote the use of `latest because I don't push such a tag to quay.io. The installation instructions refer to a versioned manifest containing the tag of a released version. The manifest in master is for local development only.

ben-st:
@awh 
since there is no latest version in the quay.io/repository/weaveworks/kured repository i thought it might be a better idea to use an existing version.

we could provide a latest version in the quay repo to ensure users don't get an error while pulling a latest version that does not exist.

but i can relate if you don't want to change your dev workflow.

greetings

Benjamin

awh:
@ben-st it's not just a matter of convenience - there are practical problems with what you're suggesting. If I were to add a new CLI option on `master` that is not in `1.0.0`, how should that be handled? I can't add it to the manifest in `master` because it won't work with the specified image. That's why I have arranged things to discourage people from installing releases using the manifest in git.

ben-st:
i was just a little confused why you don't push the latest version to the repo because this would be the way i would have handled it probably. (or maybe use -dev instead of latest only local)
But you are correct the link in the documentation is the way to use this.

I suggest we close this with the reference to the versioned .yml file, i am perfectly fine with this 👍

PR #20, opened 2018-05-15
adding option to skip rebooting during weekends

lunarfs:
We have a dev cluster that we realy do not monitor during weekends, if it blows up during the weekend (it sometimes does that) it can be in a pretty bad state when we get into work monday. adding this option to wait with rebooting until we are in the office (or close to) should make it a bit less work

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ Flags:
       --reboot-sentinel string      path to file whose existence signals need to reboot (default "/var/run/reboot-required")
       --slack-hook-url string       slack hook URL for reboot notfications
       --slack-username string       slack username for reboot notfications (default "kured")
+      --no-weekends-reboot bool     option to not restart during weekends (default "false")
 ```
 
 ### Reboot Sentinel File & Period
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ Flags:
       --reboot-sentinel string      path to file whose existence signals need to reboot (default "/var/run/reboot-required")
       --slack-hook-url string       slack hook URL for reboot notfications
       --slack-username string       slack username for reboot notfications (default "kured")
-      --no-weekends-reboot bool     option to not restart during weekends (default "false")
+      --days-to-exclude []string    List of days where rebooting should be disabled e.g. 'Sunday, Monday
 ```
 
 ### Reboot Sentinel File & Period
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -26,16 +26,16 @@ var (
 	version = "unreleased"
 
 	// Command line flags
-    period          time.Duration
-    dsNamespace     string
-    dsName          string
-    lockAnnotation  string
-    prometheusURL   string
-    alertFilter     *regexp.Regexp
-    rebootSentinel  string
-    slackHookURL    string
-    slackUsername   string
-    daysToExclude   []string
+	period         time.Duration
+	dsNamespace    string
+	dsName         string
+	lockAnnotation string
+	prometheusURL  string
+	alertFilter    *regexp.Regexp
+	rebootSentinel string
+	slackHookURL   string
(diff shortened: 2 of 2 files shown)


cmd/kured/main.go
```diff
@@ -122,6 +124,11 @@ func rebootRequired() bool {
 func rebootBlocked() bool {
 	if prometheusURL != "" {
 		alertNames, err := alerts.PrometheusActiveAlerts(prometheusURL, alertFilter)
+		weekday := time.Now().Weekday()
+		if noWeekendReboot && (int(weekday) == 0 || int(weekday) == 6) {
```
bboreham: suggest `time.Sunday` and `time.Saturday`

cmd/kured/main.go
```diff
@@ -122,6 +124,11 @@ func rebootRequired() bool {
...
+			log.Warnf("Reboot blocked: day of week: %v",weekday)
```
bboreham: seems more "info" than "warning"

tgarlot:
@bboreham , @lunarfs , would that not be better to make this more generic e.g. passing a list of WeekDay to exclude via a [StringSlice](https://godoc.org/github.com/spf13/pflag#StringSlice). Something like `--days-to-exclude=Sunday, Monday`? Happy to PR if you need help.

lunarfs:
> @bboreham , @lunarfs , would that not be better to make this more generic e.g. passing a list of WeekDay to exclude via a [StringSlice](https://godoc.org/github.com/spf13/pflag#StringSlice). Something like `--days-to-exclude=Sunday, Monday`? Happy to PR if you need help.

this is an excelent idea.. i will give it a shot..

cmd/kured/main.go
lunarfs: Thanks for the relevant comment, I have updated the PR according to the comment by  @tgarlot  and skipped the magic numbers as part of that commit

cmd/kured/main.go
lunarfs: Thanks for the relevant comment, I have updated the log to info

README.md
```diff
@@ -70,6 +70,7 @@ Flags:
       --reboot-sentinel string      path to file whose existence signals need to reboot (default "/var/run/reboot-required")
       --slack-hook-url string       slack hook URL for reboot notfications
       --slack-username string       slack username for reboot notfications (default "kured")
+      --no-weekends-reboot bool     option to not restart during weekends (default "false")
```
rndstr: this needs to be updated, too (`--days-to-exclude`)

rndstr:
would also suggest to run the changed files through `gofmt` to clean up some of the indentation issues :)

README.md
lunarfs: thanks, sorry, I should have found this one myself.

README.md
```diff
@@ -79,6 +79,7 @@ Flags:
       --reboot-sentinel string              path to file whose existence signals need to reboot (default "/var/run/reboot-required")
       --slack-hook-url string               slack hook URL for reboot notfications
       --slack-username string               slack username for reboot notfications (default "kured")
+      --days-to-exclude []string    List of days where rebooting should be disabled e.g. 'Sunday, Monday
```
sylr: Excluding days is nice but it would be nicer if we could also exclude time windows. For exemple it would be nice if  production clusters were not rebooted at night and trigger on call shift alerts.

stealthybox:
@lunarfs thanks for building this -- it's a good feature.
We just merged #66 -- usage is documented here:
https://github.com/weaveworks/kured#setting-a-schedule

Can we close this duplicate implementation? Thanks

lunarfs:
Closing this as the Feature is  implemented in https://github.com/weaveworks/kured/pull/66 , thank you for implementing this, this brings us a lot closer to be running on the master.

PR #21, opened 2018-05-15
adding option to force reboot, ignoring active allerts

lunarfs:
we have a kubernetes cluster running on Azure, sometimes we see high values for softIRQ, we resolve this issue by rebooting the nodes, we want to performe this action even if active prometheus alerts exists. when we see this high values for softIRQ we touches a separate file on the host and the node should restart (despite active alerts exists in prometheus)

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -135,7 +135,7 @@ func rebootRequired() bool {
 
 func rebootBlocked() bool {
 	if forceRebootsentinelExists() {
-		log.Infof("Force reebot sentinel %v exists, force reeboting activated",forceRebootSentinel)
+		log.Infof("Force reboot sentinel %v exists, force rebooting activated",forceRebootSentinel)
 		return false
 	}
 	if prometheusURL != "" {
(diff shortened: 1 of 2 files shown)


cmd/kured/main.go
```diff
@@ -108,6 +110,18 @@ func sentinelExists() bool {
 		return false // unreachable; prevents compilation error
 	}
 }
+func forceRebootsentinelExists() bool {
```
bboreham: please run `go fmt` on the code

lunarfs: thanx for the feedback, i have now run go fmt on the code, and fixed a bug where forcereboot was only performed if reboot was also requiered

lunarfs: @awh What is your take on this , how does this fit the overall roadmap? do you eventually have other suggestions?

SerialVelocity:
@lunarfs It might be worth just supporting the word "force" inside the existing sentinel file? It looks like this covers issue #110 I just filed

evrardjp:
@SerialVelocity that might indeed be simpler, as only the reboot blocked function would need changing by checking the content of the file.

paprickar:
I would be also interested to have this merged.
@bboreham  / @awh do you think we can have that merged soon?

PR #39, opened 2018-10-30
Enter host mount namespace

awh:
Use the tools installed in the host to effect reboots, eventually facilitating the use of commands such as `needs-restart` to determine if reboots are required.

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -97,16 +97,23 @@ func newCommand(name string, arg ...string) *exec.Cmd {
 }
 
 func sentinelExists() bool {
-	_, err := os.Stat(rebootSentinel)
-	switch {
-	case err == nil:
-		return true
-	case os.IsNotExist(err):
-		return false
-	default:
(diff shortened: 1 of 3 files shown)


cmd/kured/main.go
```diff
@@ -97,16 +97,23 @@ func newCommand(name string, arg ...string) *exec.Cmd {
...
-	switch {
-	case err == nil:
-		return true
-	case os.IsNotExist(err):
-		return false
-	default:
-		log.Fatalf("Unable to determine existence of sentinel: %v", err)
-		return false // unreachable; prevents compilation error
+	// Relies on hostPID:true and privileged:true to enter host mount space
+	sentinelCmd := newCommand("/usr/bin/nsenter", "-m/proc/1/ns/mnt", "--", "/usr/bin/test", "-f", rebootSentinel)
```
gswallow: Whose job is it to remove the /var/run/reboot-required file?  Will this result in reboot loops?

PR #42, opened 2018-11-05
adding getting alerts via alertmanager, which allows user to be able …

rmartinez3:
Hello,

Will like to add getting alerts via alertmanager, in which a user is able to silence alerts via alertmanager, when a reboot is occurring. So kured wont be blocked on active alerts. Let me know what is needed for this to be merged.

Thanks

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -26,15 +26,16 @@ var (
 	version = "unreleased"
 
 	// Command line flags
-	period         time.Duration
-	dsNamespace    string
-	dsName         string
-	lockAnnotation string
-	prometheusURL  string
-	alertFilter    *regexp.Regexp
-	rebootSentinel string
-	slackHookURL   string
-	slackUsername  string
+	period          time.Duration
+	dsNamespace     string
+	dsName          string
+	lockAnnotation  string
+	alertmanagerURL string
+	prometheusURL   string
+	alertFilter     *regexp.Regexp
+	rebootSentinel  string
+	slackHookURL    string
+	slackUsername   string
 
 	// Metrics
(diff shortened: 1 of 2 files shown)


cmd/kured/main.go
```diff
@@ -141,6 +144,20 @@ func rebootBlocked() bool {
 			log.Warnf("Reboot blocked: %d active alerts: %v", count, alertNames)
 			return true
 		}
+	} else if alertmanagerURL != "" {
+		alertNames, err := alerts.AlertmanagerActiveAlerts(alertmanagerURL, alertFilter)
+		if err != nil {
+			log.Warnf("Reboot blocked: AlertManager query error: %v", err)
+			return true
+		}
+		count := len(alertNames)
```
bboreham: Do you think this section could be merged with the one above where the code is identical?

pkg/alerts/alertmanager.go
```diff
@@ -0,0 +1,45 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	activeAlertSet := make(map[string]bool)
+
+	for _, alert := range fetchedAlerts {
+		alertName := alert.Labels["alertname"]
+		if filter == nil || !filter.MatchString(string(alertName)) {
```
bboreham: From here to the end of the function seems identical to the prometheus query code - could it be extracted into a common function?

rmartinez3: I'll work on making a common setction for both of them . thanks for looking over PR .

evrardjp:
Does this need to exist? Couldn't a generic notification command do the trick?

Issue #31, opened 2018-10-01
Parameterize the reboot function in order to be able to do other actions like shutdown for instance


mvisonneau:
An interesting feature for me would be to be able to shutdown instead of rebooting the node. Running on AWS, I would like to let the ASG replace my instance once it has been shutdown.

awh:
Huh, interesting idea! Thanks @mvisonneau

PR #43, opened 2018-11-07
Parameterise reboot command

razaj92:
As per #31, it would be useful to shutdown an instance rather than just reboot, for use cases such as cycling instances in an AWS Autoscaling Groups. 

I have tried to keep it simple by just parameterising the `reboot` value as so you can run any of the `systemctl` commands instead. If you have any other suggestions let me know.

Thanks

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -81,7 +81,7 @@ func main() {
 		"alert names to ignore when checking for active alerts")
 	rootCmd.PersistentFlags().StringVar(&rebootSentinel, "reboot-sentinel", "/var/run/reboot-required",
 		"path to file whose existence signals need to run command")
-	rootCmd.PersistentFlags().StringVar(&command, "command", "reboot",
+	rootCmd.PersistentFlags().StringVar(&command, "command", "/bin/systemctl reboot",
 		"systemctl command to run when sentinel is found, default: reboot")
 
 	rootCmd.PersistentFlags().StringVar(&slackHookURL, "slack-hook-url", "",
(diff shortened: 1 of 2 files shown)


mvisonneau:
Thanks @razaj92 :)!

yogeek:
Hello @awh @mvisonneau ! This feature would be really useful (I have the exact same scenario as @razaj92), is it planned to continue the work on this PR please ?
How can I help ?

cmd/kured/main.go
```diff
@@ -261,9 +264,9 @@ func commandReboot(nodeID string) {
 	}
 
 	// Relies on hostPID:true and privileged:true to enter host mount space
-	rebootCmd := newCommand("/usr/bin/nsenter", "-m/proc/1/ns/mnt", "/bin/systemctl", "reboot")
+	rebootCmd := newCommand("/usr/bin/nsenter", "-m/proc/1/ns/mnt", "/bin/systemctl", command)
```
evrardjp: in this case, shouldn't also systemctl be part of the "command" variable? 
In that case the default command could be "/bin/systemctl reboot". Allows more flexibility for non systemd linux systems.

renaudhager:
Can someone look at this PR?

bboreham:
(a) a small conflict with master needs resolved before it will merge
(b) there is a comment from @evrardjp that the design of this feature is wrong

hairyhenderson:
@razaj92 are you able to look at this? This'd be a super useful feature 😉

cmd/kured/main.go
razaj92: Good point!

razaj92:
Hey guys 👋 

I've rebased and addressed the comment as requested 🙂 

I am no longer using this anymore myself though, so if anyone would like to test or add to this PR please feel free!

evrardjp:
I have added the topic of reboots into our next community meeting, as there are multiple approaches currently opened.
See the agenda here https://docs.google.com/document/d/1bsHTjHhqaaZ7yJnXF6W8c89UB_yn-OoSZEmDnIP34n8/edit# .

Sorry for the delay!

evrardjp:
Let's keep that one opened. If we rebase it, we can definitely improve our reboot story here.

I would prefer to see, in this PR, an additional change to expose that new feature in the helm chart.

evrardjp:
You might be interested by this PR: https://github.com/weaveworks/kured/pull/297

dholbach:

PR #45, opened 2018-11-12
Add flag to expose only the metric

Thubo:
Idk if this is really helpful, but we just want to know if a reboot is required and perform the reboot manually.

If there's anything to improve the PR, please let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ Flags:
       --period duration             reboot check period (default 1h0m0s)
       --prometheus-url string       Prometheus instance to probe for active alerts
       --reboot-sentinel string      path to file whose existence signals need to reboot (default "/var/run/reboot-required")
+      --metric-only bool            do not reboot - expose just the metric (default false)
       --slack-hook-url string       slack hook URL for reboot notfications
       --slack-username string       slack username for reboot notfications (default "kured")
 ```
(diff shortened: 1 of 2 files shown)


awh:
Hi @Thubo - I'd rather understand why you want to reboot manually, and see if we can address that - can you explain?

Thubo:
Hi @awh ,
we have applications (pods) which are rather painful to restart, because they handle state-full connections (websockets or databases). Therefore we want to be in control about the restart of the underlying nodes (e.g. trigger the reboot manually outside business-hours and monitor the system closely).
I know this is not ideal - but we still want to be notified when a reboot is required. (A custom metric would to the same job, but I like the idea of not reinventing the wheel.)

awh:
@Thubo does #46 satisfy your requirements?

Thubo:
> @Thubo does #46 satisfy your requirements?
Yes it does. However, as you stated in the PR:

> Try not to abuse this mechanism - it's better to strive for
> restartability where possible. If you do use it, make sure you set
> up a RebootRequired alert as described in the next section so that
> you can intervene manually if reboots are blocked for too long.

We probably wouldn't abuse it, but in the end we would block pretty much all our nodes due to the topology of our services/clusters. But I guess that would be fine and we are able to pretty much achieve what we are looking for.

Issue #47, opened 2018-11-22
Notifications to Microsoft Teams


tgarlot:
Hi,

I would like to use Microsoft Teams instead of Slack for drain/reboot notification. Implementation would be a lot similar as it's also using a simple webhook with a custom JSON structure.

Would you accept a PR for that? 

thanks

strowi:
would be very interested in this..

strowi:
at least for MSTeam it just works if you put the Teams-webhook for  "--slack-hook-url=..."

mmisztal1980:
What's the status of #49 ? A lot of ppl could really use it by now.

 Can it be added and the generic impl be postponed till somebody has the actual time to do it?

PeterUherek:
Hey guys, what is a state of this feature?

mirageglobe:
did anyone have MS teams webhook work for --slack-hook-url? tried using extraargs with MS teams webhook and it did not like it. no errors and did not trigger notification even thou reboot-required file was present.

dsargent3220:
Would also like to send notifications to Teams.  Tried to use the slack URL and it didn't work.  If someone knows a way to make this work, please share here.  :)

tarioch:
It works for me I have set the slack hook url to the incoming webhook url of my teams channel.

PR #49, opened 2018-11-24
Add Microsoft Teams as notification channel

tgarlot:
Fixe #47 

Tested on AKS successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ Flags:
       --reboot-sentinel string      path to file whose existence signals need to reboot (default "/var/run/reboot-required")
       --slack-hook-url string       slack hook URL for reboot notfications
       --slack-username string       slack username for reboot notfications (default "kured")
+      --teams-hook-url string       Microsoft Teams hook URL for notifications
 ```
 
 ### Reboot Sentinel File & Period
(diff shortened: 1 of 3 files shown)


bboreham:
Thanks for the PR. Seems the code is mostly the same between Slack and Teams - just the body and the url are different - so could that be abstracted out?  Bit more DRY?

Maybe there is a library somewhere that abstracts a bunch of different notification channels?  We don't really want to do all that work in this small project.

tgarlot:
Agreed. I've even put a Teams URL instead of Slack and it work the same :-) Will look a bit further to this PR, see how I can improve it.

strowi:
Very interested in this, whats the status?

bboreham:
> Maybe there is a library somewhere that abstracts a bunch of different notification channels?

#94 offers one possibility.

evrardjp:
What do you think of a generic notification system? (See above ^)

Danlewis3:
+1

evrardjp:
Awesome, thanks for the feedback!

dsargent3220:
Hey guys.  I get the desire to have a DRY, clean, and generic implementation of this, but it hasn't happened, and in the meantime this PR has sat here for over 2 years.  It would be best to implement what has already been written then do the "nice to haves" later.

evrardjp:
@dsargent3220 someone, as of today, is working on a generic implementation in  https://github.com/weaveworks/kured/pull/228. 

Sorry if it takes more time than you would like...

tgarlot:
Will close this PR as there is a better solution being worked out.

PR #56, opened 2018-12-27
Build arm64

jmreicha:
* Adds support for building and pushing arm64 Docker images
* Updates kubectl to v1.13.1

Since Quay does not currently support "fat manifests", the images are tagged the same way with the addition of the `-arm64` suffix for the new architecture.

Tested on my Kubernetes 1.13.1 Pine64 cluster.

```
time="2018-12-27T22:48:54Z" level=info msg="Kubernetes Reboot Daemon: build-arm64-114c349"
time="2018-12-27T22:48:54Z" level=info msg="Node ID: kube-node-1"
time="2018-12-27T22:48:54Z" level=info msg="Lock Annotation: kube-system/kured:weave.works/kured-node-lock"
time="2018-12-27T22:48:54Z" level=info msg="Reboot Sentinel: /var/run/reboot-required every 1h0m0s"
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
       - deploy:
           name: Build and push image
           command: |
-            docker login -u "$DOCKER_USER" -p "$DOCKER_PASS" quay.io
+            docker login -u "$DOCKER_USER" -p "$DOCKER_PASS"
             if [ -z "${CIRCLE_TAG}" ]; then
                 make publish-image
             else
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-cmd/kured/kured
+cmd/kured/kured*
 cmd/prom-active-alerts/prom-active-alerts
 vendor
-build
+build*
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,0 +1,14 @@
+FROM alpine:3.8 as builder
+RUN apk add --no-cache ca-certificates
+ADD https://storage.googleapis.com/kubernetes-release/release/v1.12.1/bin/linux/ARCH/kubectl /kubectl
+RUN chmod 0755 /kubectl
+
+FROM BASEIMAGE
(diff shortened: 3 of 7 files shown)


awh:
Would you mind removing the kubectl version update? I'd prefer to do that in a separate PR as it's also necessary to update`k8s.io/client-go` and transitive deps...

jmreicha:
@bboreham Agreed.  My first pass is basically just to get it working.  I am new to make so will need to play around with things but I should be able to dedicate some time this weekend to DRYing it.

jmreicha:
@awh I reverted to `1.12.1`, will that work or is `1.12.4` better?

awh:
1.12.1 is good thanks that's what's on current master :+1:

jmreicha:
@awh @bboreham Do either of you know what the `apk add ca-certificates` is doing in the Dockerfile?  Kured seems to work without it, wondering if it's needed at this point.

jmreicha:
I think I have worked around the ca-certificate package issue for now.

Would there be any interest in moving kured to Dockerhub to take advantage of multiarch manifests for a single Docker image?

jmreicha:
I made some updates to the PR to create a multiarch manifest on Dockerhub, since I never heard anything back.  I think this approach simplifies the build because 

It doesn't look like the repo exists on Dockerhub so will need to be created before merging.  Also, I'm not sure how to configure CircleCI with creds for pushing to Dockerhub so may need some help with that.

jmreicha:
@awh Don't want to step on any toes here but is there any way you could look at this?

jmreicha:
@bboreham @awh Ping?

billimek:
Is there anything we can do to progress this PR along?

bboreham:
Needs rebase, at least.
If it ended with a multi-arch manifest we wouldn't need the `sed` bit in the instructions.

dholbach:
It should be easier to do this now that we removed the `kubectl` binary from `kured`.

PR #64, opened 2019-03-13
Add support for telemetrics notifications from ClearLinux

thkukuk:
I don't know if this is from general interest: we are using telemetrics (https://clearlinux.org/features/telemetry) for our internal test cluster and I wrote a patch to enable kured to also send telemetrics notifications to our server for better debugging of reboot problems in the cluster.
As it uses the logrus hooks, this overhead, if compiled in, but not enabled, is nearly zero.


evrardjp:
Hey @thkukuk , I've discussed with @dholbach on twitter, and he would be happy with a generic approach. If you don't mind, I will carry this torch forward :)

dholbach:
Just a quick PSA from me:

We'd love to see you there.

PR #70, opened 2019-04-18
kured on Windows

brunhil:
Hi all, a co-worker (@carlsoncoder) and myself have been working to extend _kured_ so that it can work with Windows containers. We have made the necessary changes and would like to start the conversation with this pull request. We have a few questions on the approach we've taken for implementing this.

1. Currently we are using a username \ password to do reboots via WinRM (PowerShell remoting). Obviously this isn't super secure. What are your thoughts on this? There are options to allow communication via certificates, but this would require some additional setup to have that certificate on the nodes.
2. Currently there is a Dockerfile for Linux, we made one for Windows too, but how do we want to handle this? We simply added the OS to the Dockerfiles (Ex: Dockerfile-windows).
3. What additional work is needed for the CI Infrastructure to enable this to be building and publishing for Windows? Should we modify the _Makefile_?

Are we missing anything else to consider?


carlsoncoder:
I've been having nothing but problems getting PSRemoting to work via certificate authentication.  I would prefer the first pass to focus on just getting it to work with a username/password for now and then consider the certificate support in the future.

rhockenbury:
Keep an eye out for privileged windows containers support.  Supposedly a KEP is coming together.   Not sure of the details yet, but there's potential to sidestep the remote powershell auth.

cmd/kured/main.go
```diff
@@ -84,131 +104,90 @@ func main() {
 	}
 }
 
-// newCommand creates a new Command with stdout/stderr wired to our standard logger
-func newCommand(name string, arg ...string) *exec.Cmd {
-	cmd := exec.Command(name, arg...)
+func isWindows() bool {
```
rhockenbury: I'm not sure of the expectations from the maintainers, but IMO splitting os-specific logic into separate files tends to be cleaner.

kured-ds-windows.yaml
```diff
@@ -0,0 +1,70 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: kured
+  namespace: kube-system
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
```
rhockenbury: How would the node locking work with two damonsets?  As implemented, it would allow a windows node and linux node to reboot at the same time.  Perhaps this isn't necessarily an issue.

jdm019000:
What is the status of this thread since May?.  Is there any additional progress on Kured Windows image version working?  Is this still waiting for a KEP?  Was testing this in AKS windows node preview.

brunhil:
Hi @jdm019000 , this has fallen by the wayside for me. I haven't tested it since the initial PR and comments. I would be interested in hearing more about the KEP if @rhockenbury has any additional input

cmd/kured/main.go
evrardjp: Agreed. Even if it's ugly as hell, it's the way many go programs deal with windows. Though, I am certainly new in this ecosystem.

dholbach:
Just a quick PSA from me:

We'd love to see you there.

brunhil:
Thanks for the feedback everyone! I'll try to get the comments updated this upcoming week

PR #75, opened 2019-05-16
Support k8s 1.14

awh:
I've tested this against 1.13.6 & 1.14.1 in minikube.


awh:
Also tested on a production 1.13 cluster.

Issue #78, opened 2019-06-18
Drain timeout


sylr:
It would be great to be able to define a timeout for pod eviction.

If drain did not succeed within the timeout kured would stop trying to evict pods and release the lock.

This would be handy when pod disruption budgets does not allow eviction.

Regards.

bboreham:
Still valid

bboreham:
Is it possible to detect the difference between "drain is taking a long time" and "drain will never happen because of pod disruption budget" ?

flbla:
https://github.com/weaveworks/kured/pull/283

matthiasritter:
We'd love to see this implemented too :-)
Will this be merged?

evrardjp:
@flbla @sylr could you have a look at #341 ? I think the current implementation is covering your case. 
However, if it's not, don't hesitate to highlight what we are missing! :)

PR #79, opened 2019-06-18
Drain timeout

sylr:
Attempt at https://github.com/weaveworks/kured/issues/78

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ compatibility of one minor version between client and server:
 | master | 1.14.1  | v11.0.0          | release-1.14        | 1.13.x, 1.14.x, 1.15.x            |
 | 1.2.0  | 1.13.6  | v10.0.0          | release-1.13        | 1.12.x, 1.13.x, 1.14.x            |
 | 1.1.0  | 1.12.1  | v9.0.0           | release-1.12        | 1.11.x, 1.12.x, 1.13.x            |
-| 1.0.0  | 1.7.6   | v4.0.0           | release-1.7         | 1.6.x, 1.7.x, 1.8.x               | 
+| 1.0.0  | 1.7.6   | v4.0.0           | release-1.7         | 1.6.x, 1.7.x, 1.8.x               |
 
 See the [release notes](https://github.com/weaveworks/kured/releases)
 for specific version compatibility information, including which
@@ -79,6 +79,7 @@ Flags:
   -h, --help                                help for kured
       --lock-annotation string              annotation in which to record locking node (default "weave.works/kured-node-lock")
       --period duration                     reboot check period (default 1h0m0s)
+      --drain-timeout duration              abort drain if not succeeded within duration (default 10m0s)
       --prometheus-url string               Prometheus instance to probe for active alerts
       --reboot-days strings                 only reboot on these days (default [su,mo,tu,we,th,fr,sa])
       --reboot-sentinel string              path to file whose existence signals need to reboot (default "/var/run/reboot-required")
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -29,6 +29,7 @@ var (
 
 	// Command line flags
 	period         time.Duration
+	drainTimeout   time.Duration
 	dsNamespace    string
 	dsName         string
 	lockAnnotation string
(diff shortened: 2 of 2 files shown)


stealthybox:
@sylr Thanks for your patch. From a first glance, it looks very reasonable.

It's going to be a while before I can actually test this.

Would you be able to post some manifests that produce Pods with a PDB that reproduces the case this behavior addresses?

k8s-conformance has an Ubuntu based Vagrantfile for creating k8s clusters that we may use as a repro environent: https://github.com/stealthybox/k8s-conformance/blob/c6fa9ca602bdd7b3a26b3568e9e7471d46146016/v1.16/kubeadm/README.md

If some other users are willing to vouch for this change or post a demo with a asciinema, I'm happy to merge this.

walkamongus:
I have a little bit different take on a drain timeout -- I actually have a use-case where I want to force a reboot to occur if a drain does not succeed within a certain amount of time. Might be something where `--drain-timeout` combines with something like `--force` to abort drain and continue with reboot after a certain duration?

mzahorik:
kubectl drain has two options, grace-period and timeout.  See:

https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#drain

According to this:

https://github.com/weaveworks/kured/blob/master/cmd/kured/main.go#L235

kured is not passing those arguments, so the defaults are to use the pod grace periods (which might be unreasonably high) and 0 timeout, which is an infinite wait.

I think it'd be less confusing to have --drain-timeout and --drain-grace-period flags refer to their equivalent kubectl drain arguments and passed into newCommand().  A new flag such as --drain-hanging-timeout could be used by this code to address any edge cases where kubectl's own internal timeout logic goes wrong.

README.md
```diff
@@ -44,7 +44,7 @@ compatibility of one minor version between client and server:
 | master | 1.14.1  | v11.0.0          | release-1.14        | 1.13.x, 1.14.x, 1.15.x            |
 | 1.2.0  | 1.13.6  | v10.0.0          | release-1.13        | 1.12.x, 1.13.x, 1.14.x            |
 | 1.1.0  | 1.12.1  | v9.0.0           | release-1.12        | 1.11.x, 1.12.x, 1.13.x            |
-| 1.0.0  | 1.7.6   | v4.0.0           | release-1.7         | 1.6.x, 1.7.x, 1.8.x               | 
+| 1.0.0  | 1.7.6   | v4.0.0           | release-1.7         | 1.6.x, 1.7.x, 1.8.x               |
```
bboreham: spurious change?

cmd/kured/main.go
```diff
@@ -235,9 +238,45 @@ func drain(nodeID string) {
...
+
+	done := make(chan error)
+	go func() { done <- drainCmd.Wait() }()
+
+	select {
+	case err := <-done:
+		if err != nil {
+			log.Fatalf("Error invoking drain command: %v", err)
+		}
+	case <-time.After(drainTimeout):
```
bboreham: Please make this optional

README.md
dholbach: Has been part of 048bba446f443d694ae121def56fa2bf48c71488 (https://github.com/weaveworks/kured/blame/master/README.md#L47).

cmd/kured/main.go
evrardjp: Agreed.

dholbach:
Just a quick PSA from me:

We'd love to see you there.

sylr:
> This change could cause data loss in the environment I work in - we have some pods that can take 40 minutes to shut down. Please make it so the behaviour is the same as before, unless you specify the option.

My change wouldn't cause data loss. If the timeout is hit, kured would just stop waiting, uncordon the node and move to another one.

bboreham:
@sylr apologies, the conversation thread did cover forcing a reboot but your PR does not as you say.

Why does it call `os.Exit()`?  I don't get what that is supposed to achieve since Kubernetes will just restart it.

dholbach:
We recently put in https://github.com/weaveworks/kured/commit/1ba3acab98fa217ea65405640706c55c89be956d so it should be easier to implement now?

dholbach (changes requested):
Needs work after https://github.com/weaveworks/kured/commit/1ba3acab98fa217ea65405640706c55c89be956d and #210

flbla:
Hi,
any news about this PR? I would really like to have this parameter! :)
thanks

evrardjp:
@flbla If you're interested by this PR, can you update it? That would be lovely, as @sylr didn't update it recently.

Thank you in advance.

flbla:
@evrardjp : I did it some months ago : https://github.com/weaveworks/kured/pull/283#issuecomment-798820238

evrardjp:
Ok let's close this one here then, my bad!

PR #81, opened 2019-07-15
Add options to delay releasing of lock

### LINKED ISSUE #76: Set delay between reboots
A flag to set delays between reboots should be added. Just like `--period` for checking for sentinel file, we should have something like `--reboot-delay` to specify a time frame between reboots of the nodes which have the sentinel file. A use case I am having is a cluster containing large pods taking 30 min to load. If I let kured do its job, the nodes are restarted too quickly for the pods to recover.

azenk:
This PR adds a couple options.

1. Support for a pod selector which will be used to check for the readiness of pods after node reboot.
1. A delay in releasing the reboot lock, which is also used as the interval for checking pod readiness.

These changes should close #76.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@
 	* [Reboot Sentinel File & Period](#reboot-sentinel-file-&-period)
 	* [Blocking Reboots via Alerts](#blocking-reboots-via-alerts)
 	* [Blocking Reboots via Pods](#blocking-reboots-via-pods)
+  * [Post-Reboot Unlock Delay](#post-reboot-unlock-delay)
+  * [Blocking Lock Release via Pods](#blocking-lock-release-via-pods)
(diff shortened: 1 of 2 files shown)


bboreham:
This is quite a complicated change.
I wonder if having an alert on the not-ready state would achieve the same effect, so we wouldn't need the logic in kured?

azenk:
That might work.  I have changed jobs since opening this PR.  I think our preference was to remove that external dependency.

evrardjp:
I agree with @bboreham .

dholbach:
Just a quick PSA from me:

We'd love to see you there.

joschi36:
I would be happy to see this PR merged as I also would prefer this behavior built-in, rather than depend on Prometheus.

PR #85, opened 2019-09-01
Add —dry-run

mblaschke:
Enables dry-run where no reboot is triggered (but metrics are published).
Usefull for checking if a cluster (or part of it) needs manual reboots.


dholbach:
Just a quick PSA from me:

We'd love to see you there.

bboreham:
I think this is OK for the stated purpose of getting metrics, but thinking more broadly, if I wanted to see what `kured` would do without actually triggering a reboot I would probably do the test _after_ `rebootBlocked()` rather than at the beginning.

evrardjp:
@bboreham Fair. It doesn't really matter to me, but if you want I can push up a follow up patch that moves this a little below :p

PR #86, opened 2019-09-05
Update go.mod and go.sum after 'go mod vendor'

bboreham:
Seeing that #77 had some `go.sum` changes, I ran `go mod vendor` then `go mod tidy` with Go 1.12 and got these changes.


squaremo:
Though .. why `go mod vendor`? I would think `go mod tidy` would do the job.

bboreham:
I'm not claiming to know best practice; I just stated what I did.

PR #88, opened 2019-09-11
Tolerate CriticalAddonsOnly, NoExecute, NoSchedule taints

albertocsm:
Proposing adding a series of common taint toleration for the type of workload (DS)

PR #90, opened 2019-09-24
Fixed a bug where kured was not being scheduled on master nodes properly

ericsuhong:
Fixed a bug where kured was not being scheduled on master nodes properly

### DIFF
--- a/kured-ds.yaml
+++ b/kured-ds.yaml
@@ -24,6 +24,7 @@ spec:
       serviceAccountName: kured
       tolerations:
         - key: node-role.kubernetes.io/master
+          operator: Exists
           effect: NoSchedule
       hostPID: true # Facilitate entering the host mount namespace via init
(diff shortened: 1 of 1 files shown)


bboreham:
I'm not sure this is a bug.  What was the `value` in your taint?

#88 seems a better way to go, modulo specifying the minimum necessary set

evrardjp:
agreed with @bboreham , again.

dholbach:
Just a quick PSA from me:

We'd love to see you there.

ericsuhong:
Kured doesn't by default schedule on master nodes with current deployment.yaml.

We need to add this toleration to make kured schedule on master nodes:
````
tolerations:
        - effect: NoSchedule
          key: node-role.kubernetes.io/master
          operator: "Equal"
          value: "true"
````

PR #100, opened 2019-11-22
Upgrading to latest alpine

adusumillipraveen:
Currently Kured is being reported with having security vulnerabilities by our CSP because of the base image.


bboreham:
If they can tell us how those vulnerabilities can be exploited, that would be interesting.

However I don't wish to make a fuss.

adusumillipraveen:
Thanks @bboreham . It just scans the docker image for known vulnerabilities ( static analysis ) . So, it was detecting a high scored vulnerability CVE-2019-14697 ( raised and fixed originally in musl library). https://github.com/alpinelinux/docker-alpine/issues/34. Always good to keep the noise low on these things :)

bboreham:
This is precisely my point - it is pure noise. 
The CVE you refer to was an i386-only bug.

PR #132, opened 2020-04-08
Bump to newer golang version

evrardjp:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: cimg/go:1.13
+      - image: cimg/go:1.14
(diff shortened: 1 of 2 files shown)


dholbach:
`/k/kubernetes` uses [1.13 still](https://github.com/kubernetes/kubernetes/blob/master/go.mod#L7), so I'm not sure about moving right now. Can we push this further down the line?

evrardjp:
Yes, no urgency.

dholbach:
k8s 1.19 moved to to go 1.15, so we'll be able to bump this for the kured release (1.6.0?) that brings k8s 1.20 support.

evrardjp:
Yes, I can probably do another patch.

PR #127, opened 2020-03-27
Update to match k8s 1.18.0

dholbach:

PR #133, opened 2020-04-14
Update for 1.4.0 release

dholbach:
Tested on minikube with k8s 1.16.9, 1.17.5, 1.18.2.


dholbach:
This largely refers to https://github.com/weaveworks/kured/pull/127

I tested this in our Dev environment at Weaveworks too.

dholbach:
Looking at https://github.com/weaveworks/kured#kubernetes--os-compatibility and read "Kubernetes aims to provide forwards & backwards compatibility of one minor version between client and server" - I'm wondering should we have moved to 1.17 instead?

evrardjp:
Yes it might be safer to have updated to just 1.17, but I don't think it's a big problem.
People can use version 1.3.0 until they upgrade to a k8s 1.16/1.17, for which they would be able to use 1.3.0, upgrade to 1.4.0, and then upgrade to k8s further. That's not good enough IMO.

dholbach:
Mh. Not sure - we could try to get back to 1.17 and be clearer about what we support.

evrardjp:
I am fine with that approach too.

dholbach:
This will need changing once #135 is merged.

PR #134, opened 2020-04-20
print the node id when commanding a reboot

crgarcia12:
Before commanding a reboot, it would be good if the node id is printed.
In the cloud, this could be taken by any centralized monitoring tool, and trigger some actions.
For example: On Azure, Azure Monitor could get this log, send notifications
Another use case is being able to query this together with applications logs when troubleshooting


dholbach:
Hey @crgarcia12 - good thinking! Thanks a lot for the PR! :sunflower:

PR #141, opened 2020-05-08
Add testing using kind


evrardjp:

### DIFF
--- a/kind.sh
+++ b/kind.sh
@@ -1,0 +1,247 @@
+#!/bin/sh
+
+set -o errexit
+###############
+# User config #
+###############
+
+# How to use this?
+# Define env vars KIND_IMAGE, KIND_CLUSTER_NAME (or let them be the default),
+# and the KURED_IMAGE_DEST (the public repo which will contain the dev version
+# of the image to test).
+# The process will build kured (using make image), tag the image with
+# KURED_IMAGE_DEST, publish it somewhere, spin up a kind cluster, install
+# kured in it, create a reboot sentinel file, track if reboot is happening
+# by watching the kubernetes api.
+# The resources files/kind cluster are destroyed at the end of the process
+# unless DO_NOT_TEARDOWN is not empty
+# For CI, you should probably set -x before running this script.
+
+#Define the image used in kind, for example:
+#KIND_IMAGE=kindest/node:v1.18.2
+#KIND_IMAGE=kindest/node:v1.18.0@sha256:0e20578828edd939d25eb98496a685c76c98d54084932f76069f886ec315d694
+#KIND_IMAGE=kindest/node:v1.17.0@sha256:9512edae126da271b66b990b6fff768fbb7cd786c7d39e86bdf55906352fdf62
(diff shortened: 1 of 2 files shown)


kind.sh
```diff
@@ -0,0 +1,247 @@
...
+EOF
+}
+
+check_all_nodes_are_ready() {
+	$kubectl get nodes | grep Ready | wc -l | grep $nodecount > /dev/null
+}
+
+spinup_cluster() {
+	gen_kind_manifest
+	kind create cluster --name "${KIND_CLUSTER_NAME}" --config=$tmp_dir/kind.yaml
```
dholbach, line 120: Maybe first check if `kind` is installed?

kind.sh
```diff
@@ -0,0 +1,247 @@
...
+#KIND_IMAGE=kindest/node:v1.18.2
+#KIND_IMAGE=kindest/node:v1.18.0@sha256:0e20578828edd939d25eb98496a685c76c98d54084932f76069f886ec315d694
+#KIND_IMAGE=kindest/node:v1.17.0@sha256:9512edae126da271b66b990b6fff768fbb7cd786c7d39e86bdf55906352fdf62
+#KIND_IMAGE=kindest/node:v1.16.4@sha256:b91a2c2317a000f3a783489dfb755064177dbc3a0b2f4147d50f04825d016f55
+KIND_IMAGE=${KIND_IMAGE:-"kindest/node:v1.18.2"}
+
+# desired cluster name; default is "kured"
+KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-kured}"
+
+KURED_IMAGE_DEST=${KURED_IMAGE_DEST:-"docker.io/evrardjp/kured:latest"}
```
dholbach, line 30: Maybe complain if `KURED_IMAGE_DEST` is not set instead?

dholbach:
We could also take a look at [what Flux does](https://github.com/fluxcd/flux/tree/master/test/e2e) and steal some of their work. They use [bats](https://github.com/bats-core/bats-core) and kind for their e2e tests using GH Actions.

kind.sh
evrardjp: Ok.

kind.sh
evrardjp: ok.

evrardjp:
I had a brief look into what Flux does this morning, and very little can be reused. I will have a look at bats in more details if it would bring us more flexibility into the testing. I can see it interesting now to test the installation of kured in different ways (manifests/helm chart), and do the same functional test (and improve when we'll get more functional tests).

evrardjp:
I am not sure introducing a dependency with bats is necessary. I am refactoring this code for re-use in any case, to have a library of shell actions. I found a reliability pain point in this code, which I am now targetting.

davidkarlsen:
Could https://github.com/marketplace/actions/kind-kubernetes-in-docker-action (or other action alternatives) be of use?

dholbach:
Yep, I would think so - it's what Flux uses too.

evrardjp:
I checked that action before my long holiday :)
I eventually found out that it was less flexible for my test cases.

Because one does not prevent another, I will propose a different approach on this patch, by ensuring we have a kind config in tree, and have manual/automated (github action) testing using it. This would be flexible enough for all the cases.

evrardjp:
This is superseded by #183 .

PR #144, opened 2020-05-21
[WIP] Lock should be claimed by new owner in case TTL expired.

### LINKED ISSUE #143: When using --annotationTTL, lock is not claimed by new owner
If `Acquire()` [returns on the `ttlExpired` path](http://github.com/weaveworks/kured/blob/ec755333941a4a647ac3d782c29016e89ea540c0/pkg/daemonsetlock/daemonsetlock.go#L48), it doesn't update the annotation.

It needs to go through [this line](http://github.com/weaveworks/kured/blob/ec755333941a4a647ac3d782c29016e89ea540c0/pkg/daemonsetlock/daemonsetlock.go#L64), otherwise the lock will still show as the previous owner, and all nodes will act as if they now own it.

michalschott:
fixes #143


pkg/daemonsetlock/daemonsetlock.go
```diff
@@ -45,7 +45,9 @@ func (dsl *DaemonSetLock) Acquire(metadata interface{}, TTL time.Duration) (acqu
 			}
 
 			if ttlExpired(value.Created, value.TTL) {
-				return true, value.NodeID, nil
+				if err := dsl.Release(); err != nil {
```
bboreham, line 48: It seems suboptimal to do another read-modify-write cycle here: can't you just skip the return on line 51 if the ttl has expired?

michalschott: If I understand https://github.com/weaveworks/kured/blob/1.4.0/pkg/daemonsetlock/daemonsetlock.go#L38-L44 correctly the loop iterates through annotations and if there is one, it returns result of `value.NodeID == dsl.nodeID`.

Not enabling this feature will keep `TTL=0`, which means `ttlExpired` will always return `false` and return properly if needed. Otherwise, it will skip to [this line](https://github.com/weaveworks/kured/blob/ec755333941a4a647ac3d782c29016e89ea540c0/pkg/daemonsetlock/daemonsetlock.go#L64).

pkg/daemonsetlock/daemonsetlock.go
```diff
@@ -75,27 +73,23 @@ func (dsl *DaemonSetLock) Acquire(metadata interface{}, TTL time.Duration) (acqu
...
-			return false, err
-		}
-
-		if ttlExpired(value.Created, value.TTL) {
-			return true, nil
+			return false, false, err
 		}
 
-		return value.NodeID == dsl.nodeID, nil
+		return value.NodeID == dsl.nodeID, ttlExpired(value.Created, value.TTL), nil
```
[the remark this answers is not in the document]
michalschott: Initially my approach was to combine https://github.com/weaveworks/kured/pull/144/files/ba38d035df175a1091c35ad7ee7478fb8465bac4#diff-8baa5bb15d6f049b3a0ee81e1592fa1aR213 with https://github.com/weaveworks/kured/pull/144/files/ba38d035df175a1091c35ad7ee7478fb8465bac4#diff-8baa5bb15d6f049b3a0ee81e1592fa1aR202 and return true in case I'm not holding, but it have expired. Would that be preferred over this?

ericsuhong:
Is it possible to get this PR merged into master? (after proper review). We are currently being affected by issue: https://github.com/weaveworks/kured/issues/143, and would love to have another kured release that fixes the issue.

michalschott:
For convenance running few productions with this patch applied, no issues so far.

dholbach:
AFAICS #213 is an alternate solution to #143

PR #150, opened 2020-06-20
Initial Kured Helm-Chart

ckotzbauer:
This PR migrates the kured helm-chart from the stable helm/charts repo, adds GitHub Actions for linting and publishing and made several initial changes to the chart.

## Changes
- Added GitHub Actions to lint, test and release the chart. [`4420dc8`](https://github.com/weaveworks/kured/commit/4420dc82d6f569c54d5cd44dbda1983180aa87fa)
- Added the unchanged chart from helm/charts [`a574a67`](https://github.com/weaveworks/kured/commit/a574a67c618fede0dd7bf9f9c66e0fa0630c62fd)
- Removed the `autolock` feature [`c473caa`](https://github.com/weaveworks/kured/commit/c473caafc8de783ddac486abff8dd5518d9a263c)
- Use a template-function for labels [`02bb6d6`](https://github.com/weaveworks/kured/commit/02bb6d650ef64107c07206012765b1de21f10e57)
- Fix role inconsistencies (verb, psp apigroup) [`d3ea563`](https://github.com/weaveworks/kured/commit/d3ea5639f42c3fb88ee34d5f09626f8707a5bbd9)
- Add affinity support [`ab0b5d1`](https://github.com/weaveworks/kured/commit/ab0b5d137c174a51313a103652dc5ffa0ff484a7)
- Add config-object for cli-flags [`2afa0a9`](https://github.com/weaveworks/kured/commit/2afa0a9da7c3f547d43a2bc280eca240574a83cd)
- Add a Service and ServiceMonitor [`2bb7b79`](https://github.com/weaveworks/kured/commit/2bb7b7937ed72f37e055af764ecac8b39fc89f2c)
- Rearrange default values [`d271165`](https://github.com/weaveworks/kured/commit/d271165496414dfe5c2a08afc77ce3bf56b78362)
- Updated docs [`a815867`](https://github.com/weaveworks/kured/commit/a815867584479859efa73dfe4f1d99e21f866840)

### GitHub Actions
- The action for linting and testing, is triggered for each PR which changes something in the `charts/` directory. The lint-check and a installation with default-values in a kind-cluster has to be successful.
- The release action is triggered on each tag and publishes the current chart to the `gh-pages` branch. (Secret `BOT_GITHUB_TOKEN` is needed @dholbach)

## Follow-up PRs
There are multiple things to do in follow-up PRs when this is merged and tested.
- List this repo on Helm-Hub.
- Deprecate the old chart at helm/charts.

## Follow-up Issues
[...]

PR #152, opened 2020-06-26
Revert #139

dholbach:
This is a follow-up to #150, so we can get a 1.4.x release out that will be geared towards k8s 1.17.x.

Update to latest 1.17 kubectl: 1.17.7.


kured-rbac.yaml
```diff
@@ -8,7 +8,7 @@ rules:
 # Allow kubectl to drain/uncordon
 #
 # NB: These permissions are tightly coupled to the bundled version of kubectl; the ones below
-# match https://github.com/kubernetes/kubernetes/blob/v1.18.2/staging/src/k8s.io/kubectl/pkg/cmd/drain/drain.go
```
ckotzbauer, line 11: @dholbach Small (but negligible) nit: `v1.17.5` instead `v1.17.7`

dholbach: Well-spotted!

Issue #154, opened 2020-06-29
Create 1.4.1 release


dholbach:
## Release
- [x] Smoke-test on current k8s 1.16, 1.17, 1.18
- [x] Test in Weave dev environment 
- [x] Test helm things?
- [x] Write release notes
- [x] Release
- [ ] Announce

## Post release
Deal with rest of #140

dholbach:
@ckotzbauer Anything from a Helm POV you want to test before releasing `1.4.1`?

dholbach:
Release notes could be something like:

-----

This release adds to the previous 1.4.0 release

### Features

- Adding --annotation-ttl for automatic unlock #119

### kured Helm chart

- make sure Circle CI ignores the 'gh-pages' branch #149
- Initial Kured Helm-Chart #150

We intend to take over the [kured chart from helm/charts/stable](https://github.com/helm/charts/tree/master/stable/kured). Please consult https://github.com/weaveworks/kured/tree/master/charts/kured for upgrading from the previous chart.


Thanks a lot everyone who helped with this release, in particular @bboreham, @ckotzbauer, @dholbach, @michalschott and others who helped with testing!

### Kubernetes Version Compatibility

The daemon image contains a 1.17.x k8s.io/client-go and kubectl binary for the purposes of maintaining the lock and draining worker nodes. Kubernetes aims to provide forwards & backwards compatibility of one minor version between client and server, so this should work on 1.1[678].x.

Tested in minikube on 1.16.12, 1.17.8, 1.18.5.



-------

ckotzbauer:
@dholbach No, there's nothing to test before the release. It can be tested finally, if the release is done.
I noticed that for every release we have to update those places in the chart with the release we want to publish:
- Chart.yaml (appVersion)
- values.yaml (image.tag)
- Readme (image.tag default-value)

Maybe there's a better way in the future to automate this...

dholbach:
Ok, so we need to update those values first, and maybe the main README too?

ckotzbauer:
Yes, otherwise we always release a chart with the previous version of kured. :joy:
Should I do a quick PR for those changes?
What would you change in the main README?

dholbach:
#157

PR #155, opened 2020-06-29
prepare chart-release for 1.4.1

ckotzbauer:
Ref: #154


ckotzbauer:
The CI failure can be ignored, as it's the first release and we do not have to bump the chart-version...

PR #160, opened 2020-06-30
add README explaining how to install kured from chart

dholbach:
I used https://github.com/fluxcd/charts/blob/gh-pages/index.md as inspiration - thanks Stefan!


README.md
```diff
@@ -0,0 +1,17 @@
+# Kured CD Helm Repository
```
ckotzbauer, line 1: I think the CD was from Flux here :laughing:

dholbach:
https://weaveworks.github.io/kured/ :fireworks:

Issue #167, opened 2020-07-17
[Helm chart]: Support secret params


audunsolemdal:
My team uses flux with a public repo, and we deploy Kured as a helmrelease. https://github.com/equinor/sdp-flux/blob/prod/base/infrastructure/kured.yaml#L24

Currently there doesn't seem to be a way to set slackHookUrl from a Kubernetes secret. Slack automatically invalidates slack webhooks on public Github repos, making us unable to get slack alerts with our workflow.

ckotzbauer:
Hm, that's a valid use-case I think.

The problem is, that all these settings are cli-arguments and helm is not able to inject a secret-value by itself on chart-rendering. So the only thing which might be possible is to set the slack-hook-url as environment variable of the kured daemonset, then the chart can be modified to reference an existing secret. But that's something which kured does not support, right now...

davidkarlsen:
But it is possible to refer to a ${VARIABLE} in the args, and set the the value of the variable as a reference to a key in a secret.
@audunsolemdal note that using SOPS you can encrypt variables: https://github.com/fluxcd/flux/pull/2580
That can be a generic approach to handling secrets - and then these can be under version-control too.

audunsolemdal:
> (...) and set the the value of the variable as a reference to a key in a secret.

Maybe I'm missing something, but I don't see how I can set an environment variable to the daemonset without modifying the helm chart.

@davidkarlsen 
My understanding is that you are suggesting using extraArgs to set an env variable to a secret?

But this doesn't use  the "env" field which is left unaltered by helm Chart config, making me unable to use e.g. valueFrom.secretKeyRef
https://github.com/weaveworks/kured/blob/master/charts/kured/templates/daemonset.yaml#L100

davidkarlsen:
@audunsolemdal the first part of my comment was targeted at the chart maintainer, to inject the use of a var at https://github.com/weaveworks/kured/blob/master/charts/kured/templates/daemonset.yaml#L45 and secondly add the var in the env: section, referring to an secret and key within it.

PR #169, opened 2020-07-21
Chart: Support extraEnvVars

audunsolemdal:
See issue #167

It seems to me that extraArgs does not support my use-case. 
Support setting extraEnvs so that extraArgs can refer to env variables such as $slackHookUrl

### DIFF
--- a/charts/kured/templates/daemonset.yaml
+++ b/charts/kured/templates/daemonset.yaml
@@ -106,6 +106,7 @@ spec:
                   fieldPath: spec.nodeName
             {{- if .Values.extraEnvVars }}
               {{ toYaml .Values.extraEnvVars | indent 12 }}
+            {{- end }}
       {{- with .Values.tolerations }}
       tolerations:
 {{ toYaml . | indent 8 }}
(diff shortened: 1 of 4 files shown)


charts/kured/templates/daemonset.yaml
```diff
@@ -104,6 +104,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
+            {{- if .Values.extraEnvVars }}
```
ckotzbauer, line 107: @audunsolemdal Could you please add a "end" for the "if".

ckotzbauer:
In addition to my above comment, could you please document the new option in the README and fix the failing lint-job? (version bump is missing)
Thanks!

charts/kured/templates/daemonset.yaml
davidkarlsen: I believe this has to be left adjusted when you use indent. Also see nindent

ckotzbauer:
@audunsolemdal Thanks for your changes. I tested the chart-rendering and it only worked (with values) if I used `nindent` instead of `indent`. Could you change that as well please?
But then it should be fine!

audunsolemdal:
@ckotzbauer are any more steps needed for the new chart version to be pushed to Helm Hub?

ckotzbauer:
@audunsolemdal New versions of the chart are always published alongside Kured releases.

Issue #60, opened 2019-01-22
security hardening


kfox1111:
Kured seems to have been built with the assumption that all nodes are trustworthy.

This assumption can be broken in a multitenant environment, or in an environment when a single node gets compromised.

With a quick look through of the code, there is some low hanging fruit that could potentially be fixed to tighten up security:

* The lock is implemented by tweaking the kured daemonset. But this has enough permissions edit all the daemonsets pods in the entire cluster and replace it with a privileged container of the attackers choice. This can be fixed in two ways I can think of. One is to make the lock item different, such as using a configmap. The other is to make a separate service that is in charge of issuing locks so the nodes themselves can't.

* Permissions are granted to the kured service account to drain/cordon/uncordon workload, but there can't be a restriction on which node is allowed to be acted on. So rather then using the account to drain the current load, an attacker could drain all nodes but the current node, attracting workload along with their secrets to the current node, then read them off the host who now has access to them. This too can be solved either of 2 ways. having a central service draining/uncordoning which an be pinned to a more critical node. alternately it may be possible to use the kubelets own credentials to drain itself. The noderestriction admission controller should only allow it to act upon itself with a node credential.

adhodgson1:
In addition I am wondering whether we could tighten up the privileged container access by removing some of the unneeded Linux capabilities to help reduce the scope of the container?

arianmotamedi:
There hasn't been any activity on this in a while but do we know what the minimum set of requirements for the container to run are? Ideally we shouldn't have to run it as root.

bboreham:
@kfox1111 no plans to add a separate service in this project.
Making the locked object independent of the daemonset seems like a good idea - could be a runtime parameter?

evrardjp:
I think this might be partially linked to https://github.com/weaveworks/kured/issues/25

evrardjp:
I have a PoC code that uses the NodeRestriction Admission controller [1], but sadly I had to map the host's kubelet conf and certs into the container, which means that it probably has more access than before... (secrets?). So that's probably not a good idea.

Does anyone have a better idea, that doesn't involve writing a new service/operator?

1: https://github.com/evrardjp/kured/commit/18eb857958d38895f6faa4882dc1f163b72889ba

kfox1111:
[...]

PR #172, opened 2020-08-07
Support unprivileged container

pjbgf:
Add support to run kured as a non-privileged container also restraining it from using any capability apart from `SYS_BOOT`.

Relates to https://github.com/weaveworks/kured/issues/60 and https://github.com/SUSE/skuba/issues/1237

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -7,6 +7,8 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
+	"strings"
+	"syscall"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -127,6 +129,16 @@ func newCommand(name string, arg ...string) *exec.Cmd {
 func sentinelExists() bool {
 	// Relies on hostPID:true and privileged:true to enter host mount space
 	sentinelCmd := newCommand("/usr/bin/nsenter", "-m/proc/1/ns/mnt", "--", "/usr/bin/test", "-f", rebootSentinel)
+
+	if strings.EqualFold(os.Getenv("UNPRIVILEGED"), "true") {
+		// Relies on the sentinel existing inside the same container mount namespace.
+		// Note that by default the host sentinel tends to be placed in a sensitive folder (/var/run/),
+		// mounting that as a volume is not recommended - but probably still safer than running the container as privileged.
+
+		// Other approaches could involve an external process to check the existance of the sentinel and safely place a file
+		// in a location that is safe to be mounted into this container.
+		sentinelCmd = newCommand("/usr/bin/test", "-f", rebootSentinel)
+	}
 	if err := sentinelCmd.Run(); err != nil {
 		switch err := err.(type) {
 		case *exec.ExitError:
@@ -265,9 +277,32 @@ func commandReboot(nodeID string) {
 		}
 	}
 
-	// Relies on hostPID:true and privileged:true to enter host mount space
-	rebootCmd := newCommand("/usr/bin/nsenter", "-m/proc/1/ns/mnt", "/bin/systemctl", "reboot")
-	if err := rebootCmd.Run(); err != nil {
+	var err error
+	if strings.EqualFold(os.Getenv("UNPRIVILEGED"), "true") {
+		ch := make(chan bool, 1)
+		defer close(ch)
+
+		go func() {
+			// Commits filesystem caches to disk in order to avoid data loss.
+			syscall.Sync()
+		}()
+
+		select {
+		case <-ch:
+			log.Info("sync successful")
+		case <-time.After(30 * time.Second):
+			log.Error("sync timed out")
+		}
+
+		// Relies on hostPID:true and SYS_BOOT to restart machine
+		err = syscall.Reboot(syscall.LINUX_REBOOT_CMD_RESTART)
+	} else {
+		// Relies on hostPID:true and privileged:true to enter host mount space
+		rebootCmd := newCommand("/usr/bin/nsenter", "-m/proc/1/ns/mnt", "/bin/systemctl", "reboot")
+		err = rebootCmd.Run()
+	}
+
+	if err != nil {
 		log.Fatalf("Error invoking reboot command: %v", err)
 	}
 }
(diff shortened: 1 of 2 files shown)


evrardjp:
Awesome! I will review this.
Some folks intended to make that reboot command configurable, so we might have to communicate on a few issues about it.

pjbgf:
Hey @evrardjp, 

Thank you for taking the time to review the changes, you gave me a lot of food for thought.
I do agree that flexibility is key - specially when we can keep the attack surface small.

Since your last review I made a few changes:

1) The changes to `main.go` are now backwards compatible. Meaning that users can continue to use kured just as they did before. However, users wanting to run as unprivileged containers would have that option by using  `kured-ds-unprivileged.yaml` instead.

2) To run without `privileged` an environment variable `PRIVILEGED` with value `true` will need to set.

3) I renamed the sentinel being used on the Daemonset to explicitly show the flexibility with this approach, however, they could both have the same name/location.

As for the reboot and other commands, I think those are great additions and can see the value they would add. I wonder whether they need a full branch of its own or if they could just be added as follow-up PRs? But I am happy either way. :+1: 

PTAL

cmd/kured/main.go
```diff
@@ -265,9 +277,9 @@ func commandReboot(nodeID string) {
 		}
 	}
 
-	// Relies on hostPID:true and privileged:true to enter host mount space
-	rebootCmd := newCommand("/usr/bin/nsenter", "-m/proc/1/ns/mnt", "/bin/systemctl", "reboot")
-	if err := rebootCmd.Run(); err != nil {
+	// Relies on hostPID:true and SYS_BOOT to restart machine
+	err := syscall.Reboot(syscall.LINUX_REBOOT_CMD_RESTART)
```
evrardjp, line 281: I would also wrap with the unprivileged condition, and use the systemctl reboot for privileged ones.

AFAIK, the syscall will only do the reboot, and therefore we might be at least missing a _sync_  in the reboot process, to not corrupt data... WDYT?

kured-ds-unprivileged.yaml
```diff
@@ -0,0 +1,140 @@
+---
```
saschagrunert, line 1: Is there any drawback so that we do not make this the new default?

pjbgf: I don't see any major problem on pushing this to be the default. 
The new approach won't break existing workloads and we could update the `yaml` and helm chart to be a seamless replacement. 

The only down side of the new approach is having two `daemonset`s, which in my opinion is a tiny downside when considering we will finally be able to run kured unprivileged and with a single Linux capability.

Shall we get this merged and then a follow up PR to just make it the default so we can hear more feedback just around that possibility?

saschagrunert: Yeah sounds reasonable, I tend to make it the default. 👍

cmd/kured/main.go
pjbgf: Good spot on the `sync`. We actually we need it in both cases, so I have amended my PR to include that instead.

pjbgf:
@saschagrunert @evrardjp Just added a call to `syscall.Sync()` pre-reboot to avoid data loss - which is the same process as systemctl does internally, PTAL.

evrardjp:
It's kinda the same process as systemd's shutdown target, but it's not _exactly_ the same, which leaves us with a gap IMO.

the shutdown service will do quite more things [0], if I am not mistaken (the sync can hang for example, so systemd has implemented code [1] to be more reliable). It means we'll have to do the same, as we are _effectively_ reimplementing a reboot procedure here.

[0]: https://github.com/systemd/systemd/blob/33d943d168319dcda321d4a758fba8de81c2d0d4/src/shutdown/shutdown.c#L382-L547
[1]: https://github.com/systemd/systemd/blob/33d943d168319dcda321d4a758fba8de81c2d0d4/src/shutdown/shutdown.c#L235-L237

pjbgf:
> It's kinda the same process as systemd's shutdown target, but it's not _exactly_ the same, which leaves us with a gap IMO.
> the shutdown service will do quite more things [0](https://github.com/systemd/systemd/blob/33d943d168319dcda321d4a758fba8de81c2d0d4/src/shutdown/shutdown.c#L382-L547), if I am not mistaken (the sync can hang for example, so systemd has implemented code [1](https://github.com/systemd/systemd/blob/33d943d168319dcda321d4a758fba8de81c2d0d4/src/shutdown/shutdown.c#L235-L237) to be more reliable). It means we'll have to do the same, as we are _effectively_ reimplementing a reboot procedure here.

@evrardjp Thank you for the feedback. I made a few changes to keep the existing behaviour and also add an async call to `sync()` with a timeout, which should be analogous to the systemd's implementation. This should get us the best of both worlds.

PTAL

evrardjp:
While there are still differences, I think this should be "good enough" as a first pass.
Thanks @pjbgf : )

pjbgf:
@evrardjp @saschagrunert any ideas how can we move this forward?

saschagrunert (approved):
Code LGTM, @dholbach please take a look :)

evrardjp:
@dholbach if we are worried that this code might be too disruptive, maybe we can start another branch.

ckotzbauer (approved):
Don't tested it, but the code looks good!

evrardjp:
My biggest concern is that we reimplement the wheel of a restart process, which is what systemd is already tackling. However, as said above, I think we should merge it and iterate, maybe in a different branch. I feel like the work could quite reduce the surface :)

pjbgf:
Hey @evrardjp @dholbach, any news on this one?

evrardjp:
We've discussed it in our meeting, but the tl;dr: is that we need to evaluate the different approaches for reboot in more detail.
We might want to make this bit of code an optional path instead of being the unique way. 

For that we, might want to add a flag at kured to define the method of reboot (syscall/command), to keep the current behaviour unchanged.

pjbgf:
Is it worth reopening it and implementing it based on the new flags then or is running privileged the preferred approach?

evrardjp:
AFAIK, there is still no official stance on the way forward. I rewrote bits of the reboot logic to aggregate a few PR, and thought about this one while doing so. I would say it's not worth rebasing this PR as of now, but at the same time, I would say that running unprivileged was not discarded at all.

pjbgf:
@evrardjp thank you for the update. Giving the lack of traction I will close this PR, if things change in the future we can always re-open it.

PR #180, opened 2020-08-25
Update gh actions

evrardjp:
Some actions have newer releases, we should use them.

### DIFF
--- a/.github/workflows/check-links.yaml
+++ b/.github/workflows/check-links.yaml
@@ -8,7 +8,7 @@ jobs:
     - uses: actions/checkout@v1
     - name: Link Checker
       id: lc
-      uses: peter-evans/link-checker@v1
+      uses: peter-evans/link-checker@v1.2.2
       with:
         args: -r *.md *.yaml */*/*.go -x .cluster.local
     - name: Fail if there were link errors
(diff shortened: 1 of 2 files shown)


.github/workflows/check-links.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
     - uses: actions/checkout@v1
     - name: Link Checker
       id: lc
-      uses: peter-evans/link-checker@v1
+      uses: peter-evans/link-checker@v1.2.2
```
dholbach, line 11: This shouldn't be necessary if I understand https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses correctly:

> Using the specific major action version allows you to receive critical fixes and security patches while still maintaining compatibility. It also assures that your workflow should still work.

Might be worth switching to major versions everywhere?

@ckotzbauer WDYT?

davidkarlsen:
TIP. You can automate action updates with dependabot nowadays.

.github/workflows/check-links.yaml
ckotzbauer: I would also prefer to only specify the major version. Only if it's really necessary to do otherwise.

evrardjp:
@davidkarlsen I will have a look at this. That PR doesn't really make sense anymore if that's the case.

PR #181, opened 2020-08-25
Add manifests testing

evrardjp:
This extends our test coverages for kured-* manifest changes on PRs,
and any eventual changes in kubernetes/kubectl on periodics.

Signed-off-by: Jean-Philippe Evrard <open-source@a.spamming.party>


evrardjp:
Meh, it seems the retry action is not really working. Do not merge yet.

evrardjp:
I think we would benefit from a makefile handling the versioning bits in the manifests/helm charts. I can tackle that as another patch after this.

evrardjp:
This is ready for review.

.github/workflows/kustomize-test.yml
```diff
@@ -0,0 +1,53 @@
...
+      # and therefore kubernetes will break with our manifests
+      # This creates a 1 node cluster.
+      - name: Create a default kind cluster
+        uses: helm/kind-action@master
+
+      # Kustomize won't work with symlinks to resources outside its folder
+      # kind-action already installed kind and kubectl.
+      - name: Deploy with kubectl using kustomization
+        run: |
+          cp kured-{ds,rbac}.yaml .github/kustomize/
```
evrardjp, line 33: Maybe we could have a manifests folder in the root of the project, to allow people to deploy with kustomize by default, instead of hiding it in there? WDYT?

.github/workflows/kustomize-test.yml
```diff
@@ -0,0 +1,53 @@
...
+          VERSION=$(curl -s https://api.github.com/repos/weaveworks/kured/releases | jq -r .[0].tag_name)
+          sed -i "s#docker.io/weaveworks/kured#docker.io/weaveworks/kured:$VERSION#g" .github/kustomize/kured-dev.yaml
```
evrardjp, line 35: This can be automated with the Makefile, see also https://github.com/weaveworks/kured/pull/185 .
However, if we do the #185 and we have a manifests/kustomize folder, it means that we don't need to do this step.

.github/workflows/kustomize-test.yml
ckotzbauer: :+1: for a manifest folder in the repo root

.github/workflows/kustomize-test.yml
ckotzbauer: We should go with #185 and then remove these lines :+1:

ckotzbauer:
Only question: The schedule is only to test new versions of kind?

ckotzbauer (approved):
Apart from my question, this looks good to me

evrardjp:
Now that 185 has been merged, I will refactor this.

.github/workflows/kustomize-test.yml
evrardjp: Done!

dholbach (approved):

PR #183, opened 2020-08-28
Add smoke/basic functional test

evrardjp:
Without this patch, we don't test on release whether kured actually
works and behave well.

This is a problem, as a functional issue could have been hidden by
a recent change, as our testing is minimalist (only test the
usability, not the functionality).
Instead of testing manually, we should ensure this in CI.

This fixes it by adding a github action which tests the previously
built artifacts before publishing a release. The job consume the helm
chart in our code tree  (note: this relies on the last released image),
and run a functional test triggering a coordinated restart of a
whole 5 node cluster deployed with kind, through github actions.

Note: The github action needs to reset docker configuration, else
the reboot of the node (a docker container in kind) will fail.
It will be correctly triggered, but the node will not come back up,
with its systemd log mentioning: "Failed to attach 1 to compat systemd cgroup".

### DIFF
--- a/.github/workflows/smoke-tests.yml
+++ b/.github/workflows/smoke-tests.yml
@@ -1,0 +1,66 @@
+# This needs all the previous artifacts to be created:
+# the image should be published, the repo should be tagged, the helm
+# chart pushed. It's assumed everything is working, and we are doing
+# a last final test at the release creation time to _ensure_ we can
+# release. At this point it's still time to back off in case of problem,
+# not publish the release and iterate on tags.
+
+name: Smoke test
+
+on:
+  push:
+    branches:
+      - '**'
+  release:
+    types: created
+
+jobs:
+  deploy-manifests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: "0"
+
+      - name: Workaround "Failed to attach 1 to compat systemd cgroup /actions_job/..." on gh actions
+        run: |
+          sudo bash << EOF
+              cp /etc/docker/daemon.json /etc/docker/daemon.json.old
+              echo '{}' > /etc/docker/daemon.json
+              systemctl restart docker || journalctl --no-pager -n 500
+              systemctl status docker
+          EOF
+
+      - name: Create 5 node kind cluster
+        uses: helm/kind-action@master
+        with:
+          config: .github/kind-cluster.yaml
+
+      - name: Deploy kured on default namespace with its helm chart
+        run: |
+          curl -o helm3.tgz https://get.helm.sh/helm-v3.3.0-linux-amd64.tar.gz
+          tar xf helm3.tgz
+          ./linux-amd64/helm install kured ./charts/kured/ --set configuration.period=1m
+          kubectl config set-context kind-chart-testing
+          kubectl get ds --all-namespaces
+          kubectl describe ds kured
+
+      - name: Ensure kured is ready
+        uses: nick-invision/retry@v1
+        with:
+          timeout_minutes: 10
+          max_attempts: 10
+          retry_wait_seconds: 60
+          # DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE should all be = 5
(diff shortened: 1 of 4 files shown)


.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,66 @@
...
+# the image should be published, the repo should be tagged, the helm
+# chart pushed. It's assumed everything is working, and we are doing
+# a last final test at the release creation time to _ensure_ we can
+# release. At this point it's still time to back off in case of problem,
+# not publish the release and iterate on tags.
+
+name: Smoke test
+
+on:
+  push:
```
evrardjp, line 11: we might want to remove the push bit. I think it's convenient, but it might not make a whole lot of sense.

ckotzbauer (commented):
This looks good to me in general. However, I'm not a bash expert, maybe someone else can have a final look. :wink:

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,66 @@
...
+          EOF
+
+      - name: Create 5 node kind cluster
+        uses: helm/kind-action@master
+        with:
+          config: .github/kind-cluster.yaml
+
+      - name: Deploy kured on default namespace with its helm chart
+        run: |
+          curl -o helm3.tgz https://get.helm.sh/helm-v3.3.0-linux-amd64.tar.gz
```
dholbach, line 42: Maybe update to latest?

.github/workflows/smoke-tests.yml
```diff
@@ -0,0 +1,66 @@
...
+          tar xf helm3.tgz
```
dholbach, line 43: Does this need a `z`?

tests/kind/follow-coordinated-reboot.sh
```diff
@@ -0,0 +1,76 @@
...
+    (( attempt_num++ ))
+done
+if [[ "$DEBUG" == "true" ]]; then
+    # This is useful to see if containers have crashed
+    echo "Debug logs"
+    echo "docker ps:"
+    docker ps
+    # This is useful to see if the nodes have _properly_ rebooted.
+    # It should show the reboot/two container starts per node.
+    for name in chart-testing-control-plane chart-testing-control-plane2 chart-testing-control-plane3  chart-testing-worker  chart-testing-worker2; do
```
dholbach, line 65: Are the names fixed?

.github/workflows/smoke-tests.yml
evrardjp: Sure. Will convert this to the the latest installation through script: `curl https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 | bash`.

.github/workflows/smoke-tests.yml
evrardjp: Not anymore :)

tests/kind/follow-coordinated-reboot.sh
evrardjp: They are matching the kind configuration, which contains those node names, and the name of the cluster. While I can change this, it's just convenient to not parse the kind configuration here. I could use `kind get nodes` though, if you prefer.

dholbach: No, I can see it now. This WFM.  :+1:

.github/workflows/smoke-tests.yml
dholbach: :+1:

tests/kind/follow-coordinated-reboot.sh
evrardjp: kind get nodes will not have the docker name, so it means messing up with the output (remove the load balancer, sed the outputs, etc.). I feel like this is already complex script, so I would prefer if we change that in another PR.

evrardjp: > No, I can see it now. This WFM. +1

Maybe this should be removed of the script, plain and simple. Maybe add it into another script to help test things, or put it directly in the github action. Let's tackle this in another PR.

dholbach: :+1:

dholbach (approved):

PR #185, opened 2020-08-28
Release helper

evrardjp:
This automates the manifest and helm chart version handling.
Just pass the organisation and version in the make command to
update the manifests/helm charts.

This does not automate the helm chart version and does not
create a manifest used in the release process.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 .DEFAULT: all
-.PHONY: all clean image publish-image minikube-publish
+.PHONY: all clean image publish-image minikube-publish manifest helm-chart
 
 DH_ORG=weaveworks
 VERSION=$(shell git symbolic-ref --short HEAD)-$(shell git rev-parse --short HEAD)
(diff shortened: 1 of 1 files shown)


evrardjp:
If people are okay with this principle, I can probably adapt the developer documentation.

ckotzbauer (approved):
Not tested the `sed` commands, but the principle looks good.

Makefile
```diff
@@ -33,3 +33,13 @@ publish-image: image
 
 minikube-publish: image
 	$(SUDO) docker save docker.io/$(DH_ORG)/kured | (eval $$(minikube docker-env) && docker load)
+
+manifest:
+	sed -i "s#image: docker.io/.*kured.*#image: docker.io/$(DH_ORG)/kured:$(VERSION)#g" kured-ds.yaml
```
evrardjp, line 38: Side note: I would prefer if our manifests recorded in a specific branch are complete with their version. It allows to build an image, a consuming manifest and helm chart at the same time, and publish them all.

squaremo (approved):
Looks like it'll work, thanks @evrardjp 🍎

PR #198, opened 2020-09-10
fix: Follow DKL-DI-0004 guideline

evrardjp:
Without this patch, we need to build a cache and remove it
afterwards.  Since apk allows to work with no-cache and won't
leave artifacts, we should use it.

This will make the dockle best practices scanner happier.


ckotzbauer (approved):

evrardjp:
Rebased.

Issue #124, opened 2020-03-25
use kubectl/pkg/drain instead of binary


alexeldeib:
(didn't see an existing issue)

There's been a lot of great work to extract the drain logic to https://github.com/kubernetes/kubectl/blob/master/pkg/drain/drain.go

We could take advantage of this programmatically and remove the kubectl binary from the docker image.

dholbach:
Could https://github.com/weaveworks/wksctl/pull/271 serve as example code here?

luxas:
yes :+1:

evrardjp:
@atighineanu are you interested by working on this?

PR #199, opened 2020-09-10
feat: Add security scanning into CI

evrardjp:
Without this patch, there is no way we can see, in the development
process, if the image we are about to publish is insecure.

This is a problem as we might be releasing new versions of kured
with outdated base image which contains vulnerabilities.

This fixes it by creating a job which will show any eventual
vulnerability.

### DIFF
--- a/.github/containerscan/allowedlist.yaml
+++ b/.github/containerscan/allowedlist.yaml
@@ -1,0 +1,7 @@
+general:
+  bestPracticeViolations:
+    # We violate this rule because we add kubectl from a remote location
+    # Instead of building it from source/copying it.
+    # Until we change our practices (e.g. have Dockerfile build kubectl
+    # in a multi-staged manner), we should skip this check
+    - CIS-DI-0009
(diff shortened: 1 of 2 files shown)


evrardjp:
This won't work until https://github.com/weaveworks/kured/pull/198 is merged.

ckotzbauer (approved):

.github/containerscan/allowedlist.yaml
```diff
@@ -0,0 +1,7 @@
+general:
+  bestPracticeViolations:
+    # We violate this rule because we add kubectl from a remote location
+    # Instead of building it from source/copying it.
+    # Until we change our practices (e.g. have Dockerfile build kubectl
+    # in a multi-staged manner), we should skip this check
+    - CIS-DI-0009
```
dholbach, line 7: https://github.com/weaveworks/kured/issues/124 should also fix this AFAICS.

dholbach (approved):
Cool, let's try this out.

.github/containerscan/allowedlist.yaml
evrardjp: correct, I was tackling that in parallel, but didn't get much time to work on it yet.

PR #206, opened 2020-09-28
Added support for time wrap in timewindow.Contains

### LINKED ISSUE #204: StartTime EndTime Flags don't wrap around - can block reboots
### Problem
If the startTime command flag is set to be after endTime kured can never reboot. timewindow.Contains() always returns false.

#### Potential Unit Test Showing the Problem 
You can add the following test case to `pkg/timewindow/timewindow_test.go` 
```go
	tests := []struct {
		days  string
		start string
		end   string
		loc   string
		cases []testcase
	}{
....
		{"mon,tue,wed,thu,fri", "9pm", "5am", "America/Los_Angeles", []testcase{
			{"2019/04/04 00:49 PDT", true}, // Would expect 12:49am to be between 9pm and 5am
			{"2019/04/04 22:49 PDT", true}, // Would expect 10:49pm to be between 9pm and 5am
			{"2019/04/05 08:59 PDT", false},
			{"2019/04/05 9:01 PDT", false},
			{"2019/03/31 10:00 PDT", false},
			{"2019/04/04 12:00 PDT", false},
		}},
```
And it will fail.

```bash
$ go test -timeout 30s github.com/weaveworks/kured/pkg/timewindow -run ^TestTimeWindows
--- FAIL: TestTimeWindows (0.00s)
    timewindow_test.go:64: (---MonTueWedThuFri--- between 21:00 and 05:00 America/Los_Angeles) contains (2019/04/04 00:49 PDT) didn't match expected result of true
    timewindow_test.go:64: (---MonTueWedThuFri--- between 21:00 and 05:00 America/Los_Angeles) contains (2019/04/04 22:49 PDT) didn't match expected result of true
FAIL
FAIL    github.com/weaveworks/kured/pkg/timewindow      0.006s
FAIL
```
If kured is setup this way nothing happens other timewindow.Contains() returns false for all times.

### Impact
A user of kured can improperly configure kured in way that will result in kured never rebooting for no obvious reason. Nothing is printed to the log. The startup of kured looks reasonable. (The following is for a slightly different configuration).

```bash
$ kubectl logs -n kube-system kured-hsgrx
time="2020-09-14T20:05:21Z" level=info msg="Kubernetes Reboot Daemon: 1.4.5"
time="2020-09-14T20:05:21Z" level=info msg="Node ID: <REDACTED>"
time="2020-09-14T20:05:21Z" level=info msg="Lock Annotation: kube-system/kured:weave.works/kured-node-lock"
time="2020-09-14T20:05:21Z" level=info msg="Reboot Sentinel: /var/run/reboot-required every 1h0m0s"
time="2020-09-14T20:05:21Z" level=info msg="Blocking Pod Selectors: []"
time="2020-09-14T20:05:21Z" level=info msg="Reboot on: SunMonTueWedThuFriSat between 21:00 and 05:00 America/Chicago"
time="2020-09-14T20:05:21Z" level=info msg="Force annotation cleanup after: 2h0m0s"
```

### Suggested Fix
A simple stop gap would be to include in the readme and other examples a warning that the startTime flag MUST be before the endTime flag. Wrap around is not possible.

Code fixes could include:
- Don't allow kured to start if startTime is after end time with a meaningful error message.
- Support time wrap around in timewindow.Contains.

chentex:
Time Wrap validation

As stated in the issue #204 it is not possible to time wrap rules. For that we need to add the logic that will adhere to those scenarios.

First we check for start and end time, if start is after end time, we need to validate if we want to wrap to the day before or to the day after.

For that we check the `loctime` value to see if it is before end time, we wrap with the day before, otherwise we wrap to the next day.

I added suggested test scenarios and others that check for a time wrap with not valid days.

fixes #204

### DIFF
--- a/pkg/timewindow/timewindow.go
+++ b/pkg/timewindow/timewindow.go
@@ -47,6 +47,19 @@ func (tw *TimeWindow) Contains(t time.Time) bool {
 	start := time.Date(loctime.Year(), loctime.Month(), loctime.Day(), tw.startTime.Hour(), tw.startTime.Minute(), tw.startTime.Second(), 0, tw.location)
 	end := time.Date(loctime.Year(), loctime.Month(), loctime.Day(), tw.endTime.Hour(), tw.endTime.Minute(), tw.endTime.Second(), 1e9-1, tw.location)
 
+	// Time Wrap validation
+	// First we check for start and end time, if start is after end time
+	// Next we need to validate if we want to wrap to the day before or to the day after
+	// For that we check the loctime value to see if it is before end time, we wrap with the day before
+	// Otherwise we wrap to the next day.
+	if tw.startTime.After(tw.endTime) {
+		if loctime.Before(end) {
+			start = start.Add(-24 * time.Hour)
+		} else {
+			end = end.Add(24 * time.Hour)
+		}
+	}
+
 	return (loctime.After(start) || loctime.Equal(start)) && (loctime.Before(end) || loctime.Equal(end))
 }
 
--- a/pkg/timewindow/timewindow_test.go
+++ b/pkg/timewindow/timewindow_test.go
@@ -20,12 +20,12 @@ func TestTimeWindows(t *testing.T) {
 		cases []testcase
 	}{
 		{"mon,tue,wed,thu,fri", "9am", "5pm", "America/Los_Angeles", []testcase{
-			{"2019/04/04 00:49 PDT", false},
-			{"2019/04/05 08:59 PDT", false},
-			{"2019/04/05 9:01 PDT", true},
 			{"2019/03/31 10:00 PDT", false},
+			{"2019/04/04 00:49 PDT", false},
 			{"2019/04/04 12:00 PDT", true},
 			{"2019/04/04 11:59 UTC", false},
+			{"2019/04/05 08:59 PDT", false},
+			{"2019/04/05 9:01 PDT", true},
 		}},
 		{"mon,we,fri", "10:01", "11:30am", "America/Los_Angeles", []testcase{
 			{"2019/04/05 10:30 PDT", true},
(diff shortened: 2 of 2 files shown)


pkg/timewindow/timewindow_test.go
```diff
@@ -40,6 +40,34 @@ func TestTimeWindows(t *testing.T) {
 			{"2019/04/18 00:00 UTC", true},
 			{"2019/04/18 23:59 UTC", true},
 		}},
+		{"mon,tue,wed,thu,fri", "9pm", "5am", "America/Los_Angeles", []testcase{
+			{"2019/04/04 22:49 PDT", true},
+			{"2019/04/05 00:49 PDT", true},
+			{"2019/04/04 00:49 PDT", true},
```
evrardjp, line 46: nit: I would add this line above, so that it's easy to see them "chronologically".

pkg/timewindow/timewindow_test.go
```diff
@@ -40,6 +40,34 @@ func TestTimeWindows(t *testing.T) {
...
+			{"2019/04/05 08:59 PDT", false},
+			{"2019/04/05 9:01 PDT", false},
+			{"2019/03/31 10:00 PDT", false},
+			{"2019/04/04 12:00 PDT", false},
```
evrardjp, line 50: nit: Might be worth checking if UTC still works fine with the new code for those conditions.

pkg/timewindow/timewindow_test.go
```diff
@@ -40,6 +40,34 @@ func TestTimeWindows(t *testing.T) {
 			{"2019/04/18 00:00 UTC", true},
 			{"2019/04/18 23:59 UTC", true},
 		}},
+		{"mon,tue,wed,thu,fri", "9pm", "5am", "America/Los_Angeles", []testcase{
+			{"2019/04/04 22:49 PDT", true},
+			{"2019/04/05 00:49 PDT", true},
+			{"2019/04/04 00:49 PDT", true},
+			{"2019/04/05 08:59 PDT", false},
+			{"2019/04/05 9:01 PDT", false},
+			{"2019/03/31 10:00 PDT", false},
```
evrardjp, line 49: I would also add 31/03 at 2200 to false and the 30/03 at 0400 to false in the test coverage, as those are the edge cases.

evrardjp (commented):
Unit tests are nice!

pkg/timewindow/timewindow.go
```diff
@@ -47,6 +47,19 @@ func (tw *TimeWindow) Contains(t time.Time) bool {
...
+	// For that we check the loctime value to see if it is before end time, we wrap with the day before
+	// Otherwise we wrap to the next day.
+	if tw.startTime.After(tw.endTime) {
+		if loctime.Before(end) {
+			start = start.Add(-24 * time.Hour)
+		} else {
+			end = end.Add(24 * time.Hour)
+		}
+	}
+
```
evrardjp, line 62: I would move the function to check the days _after_ this one, as this effectively manipulates them, what do you think?

evrardjp: meh nevermind.

evrardjp (approved):
LGTM

evrardjp (approved):
LGTM

dholbach:
Thanks a lot for your work on this, @chentex

PR #209, opened 2020-10-15
Include namespace for service

stuartpb:


ckotzbauer:
Hi @stuartpb,
thanks for your PR. Can you explain, why this is needed? Helm automatically adds the namespace of the release to the manifests on install...

Issue #125, opened 2020-03-27
Suggestion: Add who is sending the alert from where when using slack


jonaspetersorensen:
Example today:
> Draining node aks-nodepool1-35951703-vmss00000b
> Rebooting node aks-nodepool1-35951703-vmss00000b

When you have multiple components that send alerts to slack then messages like these turn into a guessing game for who did what where. Add more clusters into the mix and it can get quite confusing.

Suggestion:
> kured is draining node aks-nodepool1-35951703-vmss00000b in cluster please-work-10
> kured is rebooting node aks-nodepool1-35951703-vmss00000b in cluster please-work-10

adhodgson1:
You can use the --slack-username parameter to override the notification user, I have been doing this for ages with the cluster name as this parameter and it works well.

SirBrecilos:
settings slack-username does not seem to work for me
Tried setting it via configuration & extraArgs object

Any suggestions?
```
helm get values kured:
  configuration:
    slackUsername: prod
  extraArgs:
    slack-username: prod
```

DaniJG:
When using the new recommended webhook approach in slack (creating an app and adding incoming webhooks as per https://api.slack.com/messaging/webhooks), each webhook URL is statically associated with a username and a channel.
Directly from their docs

PR #212, opened 2020-10-21
GH-125, add additional parameters for the drain/reboot slack message template

DaniJG:
Adds a way to tackle #125 by adding extra parameters `--slack-message-drain` and `--slack-message-reboot` which allow users to override the default slack messages `Draining node %s` and `Rebooting node %s`.

This is particularly helpful when using the current recommended way in slack to create webhooks (https://api.slack.com/messaging/webhooks) where the username and channel are both statically linked to the webhook URL. Using these parameters users can add extra information like the cluster name, region, and whatever else they need.

Sample using minikube for testing:
Screen Shot 2020-10-20 at 8 58 59 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,8 @@ Flags:
       --slack-channel string                slack channel for reboot notfications
       --slack-hook-url string               slack hook URL for reboot notfications
       --slack-username string               slack username for reboot notfications (default "kured")
+      --slack-message-drain string          slack message template when notifying about a node being drained (default "Draining node %s")
+      --slack-message-reboot string         slack message template when notifying about a node being rebooted (default "Rebooting node %s")
       --start-time string                   only reboot after this time of day (default "0:00")
       --time-zone string                    use this timezone to calculate allowed reboot time (default "UTC")
 ```
@@ -217,6 +219,8 @@ you immediately prior to rebooting a node:
 
 We recommend setting `--slack-username` to be the name of the
 environment, e.g. `dev` or `prod`.
+
+Alternatively you can use the `--slack-message-reboot` and `--slack-message-reboot` to customize the text of the message, e.g. `"Draining node %s part of my-cluster in region xyz"`
 
 ### Overriding Lock Configuration
 
--- a/README.md
+++ b/README.md
@@ -94,8 +94,8 @@ Flags:
       --slack-channel string                slack channel for reboot notfications
       --slack-hook-url string               slack hook URL for reboot notfications
       --slack-username string               slack username for reboot notfications (default "kured")
-      --slack-message-drain string          slack message template when notifying about a node being drained (default "Draining node %s")
-      --slack-message-reboot string         slack message template when notifying about a node being rebooted (default "Rebooting node %s")
+      --message-template-drain string       message template used to notify about a node being drained (default "Draining node %s")
+      --message-template-reboot string      message template used to notify about a node being rebooted (default "Rebooting node %s")
       --start-time string                   only reboot after this time of day (default "0:00")
       --time-zone string                    use this timezone to calculate allowed reboot time (default "UTC")
 ```
@@ -220,7 +220,10 @@ you immediately prior to rebooting a node:
 We recommend setting `--slack-username` to be the name of the
 environment, e.g. `dev` or `prod`.
 
-Alternatively you can use the `--slack-message-reboot` and `--slack-message-reboot` to customize the text of the message, e.g. `"Draining node %s part of my-cluster in region xyz"`
+Alternatively you can use the `--message-template-drain` and `--message-template-reboot` to customize the text of the message, e.g.
+```
+--message-template-drain="Draining node %s part of *my-cluster* in region *xyz*"
+```
 
 ### Overriding Lock Configuration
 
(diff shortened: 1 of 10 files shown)


DaniJG:
I've seen the conversations about including Ms Teams and generalizing the notifications so its not tied to slack. 

I would be happy to create these parameters as `--message-drain` and `--message-reboot` instead, so they are more aligned with that direction.

pkg/notifications/slack/slack.go
```diff
@@ -43,10 +43,10 @@ func notify(hookURL, username, channel, message string) error {
 	return nil
 }
 
-func NotifyDrain(hookURL, username, channel, nodeID string) error {
-	return notify(hookURL, username, channel, fmt.Sprintf("Draining node %s", nodeID))
+func NotifyDrain(hookURL, username, channel, messageTemplate, nodeID string) error {
```
DaniJG, line 46: These 2 functions are now identical. I Didn't want to change the API as part of this PR, but I am also happy to refactor. For example:

- I could move `fmt.Sprintf(messageTemplate, nodeID)` to the place in main.go where the notification is sent, leaving in this file a single function `Notify`.
- I could also add a struct to this package, which will contain all the slack options (URL, channel, user, message templates). The struct would be created once from main (similar to `DaemonSetLock`). The methods `NotifyDrain` and `NotifyReboot` will be added tot he struct and only need to receive the `nodeID`.

The second refactor is a bit more involved but feels like a step in the direction of a more flexible/generic notifications approach.

README.md
```diff
@@ -94,6 +94,8 @@ Flags:
       --slack-channel string                slack channel for reboot notfications
       --slack-hook-url string               slack hook URL for reboot notfications
       --slack-username string               slack username for reboot notfications (default "kured")
+      --slack-message-drain string          slack message template when notifying about a node being drained (default "Draining node %s")
```
evrardjp, line 97: As you suggested, I think it would be better to have it independent of "slack", as many seem eager to have different notification systems (there is even a PR starting the work for that)

pkg/notifications/slack/slack.go
evrardjp: You will need someone with more experience than me in this project to review whether this API was used outside the current code.

IMO, we should probably get rid of it.

Nevertheless, this will probably be modified if we go to a generic notification system, which will probably remove a bunch of code, including this one. So I wouldn't tackle this abstraction in here, and would keep the current code. At best, I would indeed expose notify as `Notify` and use it in main, but I don't see it as a necessity myself.

I would welcome you to join the effort of the generic notification system refactoring though, your review could be extremely valuable! :)

DaniJG: yeah makes total sense to leave further changes outside of this PR and tackle as a proper refactoring!

DaniJG:
@evrardjp renamed the template parameters as discussed (so they are not tied to slack) and will leave the notification API as it is for the purposes of this PR. Let me know if you need anything else to merge!

Re making notifications more generic, is [this the current PR](https://github.com/weaveworks/kured/pull/228) you mentioned on those efforts? 
If so, adding [shouterrr](https://containrrr.dev/shoutrrr/) seems like a good idea. Once that PR is also merged, a natural step would be to replace the current slackWebhookUrl with a new list parameter where users can provide an arbitrary number of _shouterrr destinations_. All of a sudden kured would supports every notification service that [shouterr supports](https://containrrr.dev/shoutrrr/services/overview/).
- the one caveat I see is that shouterr would break the current slack username/channel parameters, since it is designed for the new style of slack webhooks where the webhook is tied to a specific slack app and channel. Therefore, the slack username/channel parameters wouldnt have any effect anymore, which might break some folks relying on them
- Kured could either assume that breaking change or leave a "legacy" slack notifier

evrardjp:
Yes this was the one. 

I thought that we could validate the CLI arguments, and slowly deprecate the slack Webhook url and other slack parameters.
In the first time, we could support two ways to configure the notifications: one with the shoutrrr style url, the other with the existing CLI args. We would simply parse the existing arguments, and build a shoutrrr url from it, if shoutrrr style url is not provided.

Not sure about building it as a list parameter for an arbitrary number of shoutrrr destinations. What's the use case there?
Can it be done in a follow up commit?

DaniJG:
Since shouterr [supports multiple destinations](https://github.com/containrrr/shoutrrr/#using-a-sender-with-multiple-urls) out of the box, a list parameter would fit in nicely and would cover anyone who needs to send a notification through slack & email or some similar combination.

In any case, I would agree the list parameter can be added as the next step from #228. 

Note the way that PR modifies the code, it will break the existing functionality for slack channel and user name. If the intention is to slowly deprecate those parameters, then it might make sense to leave the legacy code and use it when channel/user name are provided, otherwise use shouterr. After a few more releases, then remove the parameters and legacy code.

evrardjp:
Yes I thought of that approach too. Let's discuss the best way forward in that PR :)

evrardjp:
For this PR, I consider it good.

@ckotzbauer could you have a look at the helm chart, tell us what you think ?

evrardjp:
Can we rebase this?

DaniJG:
I've just rebased the latest changes from master

dholbach:
@ckotzbauer What do you think about the Helm bits?

Thanks a lot for your work on this @DaniJG - this generally looks good to me!

ckotzbauer:
The Helm related changes are looking good :+1:

dholbach (approved):

PR #119, opened 2020-03-09
Adding --annotation-ttl for automatic unlock

michalschott:
I have observed this behavior when running kured together with cluster autoscaler, reason why this idea was born.

- CA adds nodes (CoreOS in my case)
- Nodes are starting to reboot themselves after fetching updates, ie. annotation has been added (lock owned by node 123.123.123.123)
- CA decides to downscale ASG (not all nodes had a chance to reboot yet), and it decides to cut off node 123.123.123.123
- Need to manually take off annotation

This PR introduces `--annotationTTL` parameter, by default set to 0 (means do not use this feature), which allows other kured pods to take off the lock if it has expired.

Comparing kured to [coreos-update-operator](https://github.com/coreos/container-linux-update-operator) latter one is adding lock annotation onto node object which makes such feature as lockTTL completely not required.

PR #54, opened 2018-12-12
Added expiration capabilities for the lock

mvisonneau:
This new feature allow us to set an expiration date on the lock. This is particularly useful when used with `poweroff` instead of `reboot` command that this PR enables : https://github.com/weaveworks/kured/pull/43

PR #213, opened 2020-10-22
Replaced --annotationTTL with --lockTTL and made it work correctly

mvisonneau:
This is a follow-up on #54, #119 & #143

- I renamed the --annotationTTL flag into --lockTTL as I believe it is a more pertinent nomenclature
- I fixed its implementation as reported not being functional in #143

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -33,6 +33,7 @@ var (
 	dsNamespace    string
 	dsName         string
 	lockAnnotation string
+	lockTTL        time.Duration
 	prometheusURL  string
 	alertFilter    *regexp.Regexp
 	rebootSentinel string
@@ -45,8 +46,6 @@ var (
 	rebootStart string
 	rebootEnd   string
 	timezone    string
-
-	annotationTTL time.Duration
 
 	// Metrics
 	rebootRequiredGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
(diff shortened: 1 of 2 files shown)


renaudhager:
I have tested this, works as expected.

bboreham:
Thanks for the PR.

What are your thoughts regarding it being a breaking change, i.e. anyone who turned on the feature from #54 will now get a hard failure?

bboreham:
I believe this PR changes the semantics from "lock expires at a time set by the locker" to "lock expires when kured thinks it should".
Is that what you intended?  This seems more dangerous.
See discussion at https://github.com/weaveworks/kured/pull/119#discussion_r419966865

mvisonneau:
yes indeed, I believe that this approach allows more flexibility towards changes around the TTL configuration. Specially coming for either a disabled TTL or very long one which would involve a manual edition of the annotation. But I understand that this can lead to conflicts. Happy to revert this part if you feel more comfortable keeping the info on the lock itself.

bboreham:
To me it's less flexible if only one program controls the TTL.  Locks are taken by at least two programs - kured and kubectl.

mvisonneau:
I am not sure to understand what you mean @bboreham? 🤔

bboreham:
Any particular part?  We instruct users to take locks using `kubectl` at https://github.com/weaveworks/kured#disabling-reboots

mvisonneau:
@bboreham indeed thanks, my former changes would have broken this behaviour. I changed the logic around that point back to what is currently in place 👍

bboreham:
I think `Test()` is now wrong in the case that we already hold the lock and it has not expired.

Still looking for your thoughts regarding it being a breaking change, i.e. anyone who turned on the feature from #54 will now get a hard failure.

mvisonneau:
Shouldn't `Test()` return `true` if the lock is being held by the node and not expired yet? 🤔 

#54 never got merge 🤔

bboreham:
OK now I don't know what I was thinking about `Test()`. Sorry.

And I meant #119.

mvisonneau:
ah ok! Well given that #119 is not functional I believe it is fine to take this opportunity to have a more pertinent nomenclature. 🤷

bboreham:
OK, given nobody else noted a concern, let's merge this.

dholbach:
Could it be that the chart needs an update for this?

bboreham:
Nobody should have put the previous one in a chart because it was found to not work about a day after merging it.

PR #215, opened 2020-10-29
Make go lint on cmd folder happier

evrardjp:
This does a few non important cleanups for readability.
It also removes the prom-active-alerts command which is not used,
not tested, and currently broken.


evrardjp:
Added bboreham for the review, more particularily because of the removal of prom-active-alerts which we discussed in the kured slack.

evrardjp:
Disclaimer: This has not been tested using my CI tooling. Not that I expect any failure from this, though.

bboreham:
Code changes look good, thanks.
I would have mentioned the `v1` change in the commit message, in case someone is trying to understand it in a few years.
Suggest removing `prom-active-alerts` from `.gitignore` and anywhere else it is referenced.

evrardjp:
Will do, and adapt the commit message.

evrardjp:
I have split into two commits, so it's easier to see what is what.

bboreham (approved):
Excellent work, thank you!

Issue #223, opened 2020-11-05
Introduce an expiry on the issue list


evrardjp:
I think we should have a bot that automatically expires old issues.

Most of them are not gonna be tackled if there is no focus, so I think it's fair to close them, and focus only the fresh requests.

PR #224, opened 2020-11-05
Auto expire issues and PRs

evrardjp:
Without this patch, we might hold old issues and PR for a long
time. Instead we should close them. People can reopen if necessary.

This would show that we have a proper triage process, and a proper
way to handle those.

Closes: #223


ckotzbauer (approved):

dholbach (approved):
Ok, let's try this out.

Issue #117, opened 2020-03-03
Clean up notifications system and implement generic notifications


evrardjp:
Hello folks,

I've noticed that there are multiple PRs for notifications (see [1] and [2]), and I wonder if we shouldn't instead just write a generic notification system. This would avoid proliferation of ``<insert my favorite notification tool here>``, at the expense of not having "everything bundled" into kured directly.

I would like to have feedback before starting to write this.

The idea would be to remove the arguments slack-hook-url, slack-username, and introduce a generic "notification-cmd" cli argument. The argument could be a string which can be used to call a command. That command would replace ``%{severity}``, ``%{node}``, ``%{msg}``, ``%{action}`` by severity, nodeID, message, and action taken by cured.

This allows a very simple replacement, which works for everyone, whether it's a command or a webservice. This is disruptive for current slack notifications users: Slack users would need a client which handles the notifications in the container. This could be delivered by the kured community separately.

I thought this was a very clean separation of concerns: One core code, and one notification client.

@dholbach what's your opinion?

[1]: https://github.com/weaveworks/kured/pull/64
[2]: https://github.com/weaveworks/kured/pull/49

evrardjp:
That would also help on https://github.com/weaveworks/kured/issues/47#issuecomment-537041883 , on https://github.com/weaveworks/kured/pull/49 .

simskij:
Hey,

[Shoutrrr](https://github.com/containrrr/shoutrrr) maintainer here.

Let me know if I can be of any use, if you ever decide to go down the route suggested in #94. The aim of the shoutrrr project is to provide something similar to [caronc/apprise](https://github.com/caronc/apprise), but in go, allowing projects to focus on the things that add value for them, rather than duplicating efforts on solved problems.

Thanks 🙏🏼

atighineanu:
> Hey,
> 
> [Shoutrrr](https://github.com/containrrr/shoutrrr) maintainer here.
> 
> Let me know if I can be of any use, if you ever decide to go down the route suggested in #94. The aim of the shoutrrr project is to provide something similar to [caronc/apprise](https://github.com/caronc/apprise), but in go, allowing projects to focus on the things that add value for them, rather than duplicating efforts on solved problems.
> 
> Thanks 🙏🏼

Tried it. It looks indeed nice. 
https://github.com/weaveworks/kured/pull/228

Issue #94, opened 2019-11-05
Feature request:  shoutrrr implementation ( go notifications library )


lukasmrtvy:
I would be nice to implement notification library like https://github.com/containrrr/shoutrrr, to support out of the box multiple providers.

Its not so good as apprise, but there is chance that multiple providers will starts to popup soon.

PR #228, opened 2020-11-11
Implement universal notification mechanism using shoutrrr

atighineanu:
Closes https://github.com/weaveworks/kured/issues/117
Closes https://github.com/weaveworks/kured/issues/94
This PR doesn't interfere with the slack-notification logic that is in place.
Instead of using only slack-related args, one can pass to kured flags:
- --notify-url=<URL link that follows this syntax as f(chat tech solution): https://containrrr.dev/shoutrrr/services/overview/>

So far, I've tested it on SUSE CaaSP 4.2 with k8s 1.17, on vmware.  I've tested rocketchat. It worked as it should. Further this functionality can be more ellaborated, specifying (may be depending on a --verbosity  input) for each situation (like draining, cordoning, reboot and so on).

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,6 +225,22 @@ Alternatively you can use the `--message-template-drain` and `--message-template
 --message-template-drain="Draining node %s part of *my-cluster* in region *xyz*"
 ```
 
+### Notifications 
+
+When you specify a formatted URL using `--notify-url`, kured will notify
+about draining and rebooting nodes across a list of technologies. Here is 
+the syntax:
+
+slack:           `slack://tokenA/tokenB/tokenC`
+
+rocketchat:      `rocketchat://[username@]rocketchat-host/token[/channel|@recipient]`
+
+teams:           `teams://token-a/token-b/token-c`
+
+Email:           `smtp://username:password@host:port/?fromAddress=fromAddress&toAddresses=recipient1[,recipient2,...]`
+
+More details here: https://github.com/containrrr/shoutrrr/blob/master/docs/services/overview.md
+
 ### Overriding Lock Configuration
 
 The `--ds-name` and `--ds-namespace` arguments should match the name and
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -25,6 +25,7 @@ import (
 	"github.com/weaveworks/kured/pkg/delaytick"
 	"github.com/weaveworks/kured/pkg/notifications/slack"
 	"github.com/weaveworks/kured/pkg/timewindow"
+	shoutrrr "github.com/containrrr/shoutrrr"
 )
 
 var (
@@ -41,11 +42,11 @@ var (
 	rebootSentinel         string
 	slackHookURL           string
 	slackUsername          string
+	notifyURL              string
 	slackChannel           string
 	messageTemplateDrain   string
 	messageTemplateReboot  string
 	podSelectors           []string
-
 	rebootDays  []string
 	rebootStart string
 	rebootEnd   string
@@ -67,6 +68,7 @@ func main() {
 	rootCmd := &cobra.Command{
 		Use:   "kured",
 		Short: "Kubernetes Reboot Daemon",
+		PreRun: flagCheck,
 		Run:   root}
 
 	rootCmd.PersistentFlags().DurationVar(&period, "period", time.Minute*60,
@@ -92,6 +94,8 @@ func main() {
 		"slack username for reboot notfications")
 	rootCmd.PersistentFlags().StringVar(&slackChannel, "slack-channel", "",
 		"slack channel for reboot notfications")
+	rootCmd.PersistentFlags().StringVar(&notifyURL, "notify-url", "",
+		"notify URL for reboot notfications")
 	rootCmd.PersistentFlags().StringVar(&messageTemplateDrain, "message-template-drain", "Draining node %s",
 		"message template used to notify about a node being drained")
 	rootCmd.PersistentFlags().StringVar(&messageTemplateReboot, "message-template-reboot", "Rebooting node %s",
@@ -112,6 +116,18 @@ func main() {
 	if err := rootCmd.Execute(); err != nil {
 		log.Fatal(err)
 	}
+}
+
+// temporary func that checks for deprecated slack-notification-related flags
+func flagCheck(cmd *cobra.Command, args []string) {
+	if slackHookURL != "" && notifyURL != "" {
+		log.Warnf("Cannot use both --notify-url and --slack-hook-url flags. Kured will use --notify-url flag only...")
+		slackHookURL = ""
+	}
+	if slackChannel != "" || slackHookURL != "" || slackUsername != "" {
+		log.Warnf("Deprecated flag(s). Please use --notify-url flag instead.")
+	}
+
 }
 
 // newCommand creates a new Command with stdout/stderr wired to our standard logger
@@ -247,6 +263,12 @@ func drain(client *kubernetes.Clientset, node *v1.Node) {
 			log.Warnf("Error notifying slack: %v", err)
 		}
 	}
+	if notifyURL != "" {
+		if err := shoutrrr.Send(notifyURL, fmt.Sprintf(messageTemplateDrain, nodename)); err != nil {
+			log.Warnf("Error notifying: %v", err)
+		}
+	}
+
 
 	drainer := &kubectldrain.Helper{
 		Client:              client,
@@ -285,6 +307,11 @@ func commandReboot(nodeID string) {
 	if slackHookURL != "" {
 		if err := slack.NotifyReboot(slackHookURL, slackUsername, slackChannel, messageTemplateReboot, nodeID); err != nil {
 			log.Warnf("Error notifying slack: %v", err)
+		}
+	}
+	if notifyURL != "" {
+		if err := shoutrrr.Send(notifyURL, fmt.Sprintf(messageTemplateReboot, nodeID)); err != nil {
+			log.Warnf("Error notifying: %v", err)
 		}
 	}
 


cmd/kured/main.go
```diff
@@ -90,6 +93,11 @@ func main() {
 	rootCmd.PersistentFlags().StringVar(&slackChannel, "slack-channel", "",
 		"slack channel for reboot notfications")
 
+	rootCmd.PersistentFlags().StringVar(&notifyURL, "notify-url", "",
+		"a formatted url to notify accross chat types")
+	rootCmd.PersistentFlags().StringVar(&notifyMessage, "notify-message", "",
```
evrardjp, line 98: I don't think we should expose a custom notify message yet. I think it is relevant for another (follow-up) PR.

cmd/kured/main.go
```diff
@@ -90,6 +93,11 @@ func main() {
 	rootCmd.PersistentFlags().StringVar(&slackChannel, "slack-channel", "",
 		"slack channel for reboot notfications")
 
+	rootCmd.PersistentFlags().StringVar(&notifyURL, "notify-url", "",
```
evrardjp, line 96: Why do you need to expose this?
Isn't it enough to have slack-hook-url, username, and channel?

If you still want to have this notify-url part of this commit, I suggest that you handle the case where both slack-* arguments and that notify-url is provided.

cmd/kured/main.go
```diff
@@ -245,6 +253,12 @@ func drain(client *kubernetes.Clientset, node *v1.Node) {
 		}
 	}
 
+	if notifyURL != "" {
```
evrardjp, line 256: For me there should be no duplication of slackHookURL.

We need something to parse the CLI arguments earlier in the code, and a single way to send notifications instead.
Retro compatibility should come from parsing the arguments and converting those to the new mechanism instead.

evrardjp (changes requested):
Thank you for your PR! It's a good step forward!

Here are a few comments:
- Can you squash those commits, they represent one functional unit of work, for now?
- Is it normal that the go.mod wasn't changed?
- Could you refer here why you are starting to implement this?

For that last point, I remember there were issues about different notification mechanisms. Therefore, I think  we should have a `Closes: <number>` in the PR text, to ensure we'll close all those issues. 

See also my comments in the code.

cmd/kured/main.go
evrardjp: Also, if you can _remove_ code that would be deprecated by that change, that would be awesome

cmd/kured/main.go
atighineanu: Removed. Thanks for the feedback.

cmd/kured/main.go
atighineanu: Removed.

cmd/kured/main.go
atighineanu: Removed.

atighineanu:
I believe all the remarks were addressed accordingly. 
In here: https://github.com/weaveworks/kured/pull/228/files#diff-186e20195cd0d811159bcca27888fe0bdc42c0a97623c93f2ff621f03f0426a6R25
I've added the temporary var tech := slack. In the future, I think it would be nicer to whether change the hookURL syntax or to pass as an additional variable "tech" or "chat" to the function inside slack.go

cmd/kured/main.go
evrardjp: You didn't need to _remove_ the notifyURL. For me that's perfectly fine to expose the generic mechanism in a commit on this PR.

DaniJG:
Taking the conversation from #212 
Note the current changes in this PR will break the current functionality of the slack channel and user name parameters (since these parameters would now be ignored).

I would suggest a plan like:
- update the changes in this PR in order to keep the old `Notify` implementation, which should be used in case of either channel/slack username are provided. (if not provided, then use the new shouterrr implementation)
- On a follow up PR, all the current slack parameters can be marked as deprecated and a new list parameter like `--notification-urls` can be added. The logic to use either the shouterrr or legacy implementation of the notifications can then be based on whether any of the deprecated slack parameters are provided.
- Finally, after a number of releases a new PR can remove the deprecated parameters and legacy `Notify` implementation

atighineanu:
For the input in the yaml like:
```
        image: shaklotan/kured-dev:trial33  (current build)
            ...
          command:
            - /usr/bin/kured
            - --period=30s
            - --notify-url=slack://tokenA/tokenB/tokenC
            - --slack-hook-url=slack://tokenA/tokenB/tokenC
```
I've got:

```
user@localhost:~> kubectl logs pod/kured-d8j68 -n kube-system
time="2020-11-25T21:57:24Z" level=info msg="Kubernetes Reboot Daemon: master-932f91b"
time="2020-11-25T21:57:24Z" level=warning msg="Cannot use both --notify-url and --slack-hook-url flags. Kured will use --notify-url flag only..."
time="2020-11-25T21:57:24Z" level=warning msg="Deprecated flag(s). Please use --notify-url flag instead."
time="2020-11-25T21:57:24Z" level=info msg="Node ID: kurediya-worker2"
time="2020-11-25T21:57:24Z" level=info msg="Lock Annotation: kube-system/kured:weave.works/kured-node-lock"
time="2020-11-25T21:57:24Z" level=info msg="Reboot Sentinel: /var/run/reboot-required every 30s"
time="2020-11-25T21:57:24Z" level=info msg="Blocking Pod Selectors: []"
time="2020-11-25T21:57:24Z" level=info msg="Reboot on: SunMonTueWedThuFriSat between 00:00 and 23:59 UTC"
time="2020-11-25T21:57:24Z" level=info msg="Force annotation cleanup disabled."
time="2020-11-25T21:57:46Z" level=info msg="Reboot not required"
time="2020-11-25T21:58:16Z" level=info msg="Reboot not required"
time="2020-11-25T21:58:46Z" level=info msg="Reboot not required"
time="2020-11-25T21:59:16Z" level=info msg="Reboot not required"
time="2020-11-25T21:59:46Z" level=info msg="Reboot required"
time="2020-11-25T21:59:46Z" level=info msg="Acquired reboot lock"
time="2020-11-25T21:59:46Z" level=info msg="Draining node kurediya-worker2"
time="2020-11-25T21:59:46Z" level=warning msg="Error notifying: unknown service scheme for URL ''"
WARNING: ignoring DaemonSet-managed Pods: kube-system/kindnet-6jfpk, kube-system/kube-proxy-vvcgc, kube-system/kured-d8j68
time="2020-11-25T21:59:46Z" level=info msg="Commanding reboot for node: kurediya-worker2"
time="2020-11-25T21:59:46Z" level=warning msg="Error notifying: unknown service scheme for URL ''"
time="2020-11-25T21:59:46Z" level=warning msg="Failed to connect to bus: No such file or directory" cmd=/usr/bin/nsenter std=err
time="2020-11-25T21:59:46Z" level=info msg="Waiting for reboot"
```

README.md
```diff
@@ -217,6 +217,18 @@ you immediately prior to rebooting a node:
 We recommend setting `--slack-username` to be the name of the
 environment, e.g. `dev` or `prod`.
 
+### Generinc Notifications
```
evrardjp, line 220: I would just call it "Notifications".

When adding the `--slack*` deprecation, we can then remove the whole `Slack Notifications` section, so that people are not enclined to use it anymore.

cmd/kured/main.go
```diff
@@ -110,6 +115,16 @@ func main() {
...
 
+func flagcheck(cmd *cobra.Command, args []string) {
+	if slackHookURL != "" && notifyURL != "" {
+		log.Warnf("Cannot use both --notify-url and --slack-hook-url flags. Kured will use --notify-url flag only...")
+		slackHookURL = ""
+	}
+	if slackChannel != "" || slackHookURL != "" || slackUsername != "" {
+		log.Warnf("Deprecated flag(s). Please use --notify-url flag instead.")
+	}
+}
```
evrardjp, line 126: This looks good!

nit: I would also have an extra check (something we should have had in the past):
check, if `slack-hook-url` is set, that `slack-channel` is also set. Not necessary for this PR though.

cmd/kured/main.go
```diff
@@ -236,13 +251,16 @@ func release(lock *daemonsetlock.DaemonSetLock) {
...
-
 	if slackHookURL != "" {
 		if err := slack.NotifyDrain(slackHookURL, slackUsername, slackChannel, nodename); err != nil {
 			log.Warnf("Error notifying slack: %v", err)
 		}
+	} 
+	if notifyURL != "" {
+		if err := shoutrrr.Send(notifyURL, fmt.Sprintf("Draining node %s", nodename)); err != nil {
+			log.Warnf("Error notifying: %v", err)
+		}
```
evrardjp, line 263: This will be conflicting with the other PR which allow user-defined text for notifications. We'll need to keep it in mind.

dholbach:
Unfortunately after merging #212 this will require a small update.

evrardjp (commented):
THis is a good first step.

I think we'll need another PR to actively convert existing cli arguments into shoutrrr calls.

evrardjp:
> Unfortunately after merging #212 this will require a small update.

Ahah yes, that's what I meant in my review, which apparently was done in parallel of this comment.

README.md
atighineanu: Done

atighineanu:
Tested it again:
for kured yaml input:
```
      image: shaklotan/kured-dev:trial36
        ...
          command:
            - /usr/bin/kured
            - --period=30s
            - --notify-url=slack://tokenA/tokenB/tokenC
            - --slack-hook-url=slack://tokenA/tokenB/tokenC
```
I got:
```
user@localhost:~> kubectl logs pod/kured-t9zgr -n kube-system
time="2020-11-26T08:31:12Z" level=warning msg="Cannot use both --notify-url and --slack-hook-url flags. Kured will use --notify-url flag only..."
time="2020-11-26T08:31:12Z" level=warning msg="Deprecated flag(s). Please use --notify-url flag instead."
time="2020-11-26T08:31:12Z" level=info msg="Kubernetes Reboot Daemon: master-8ad6bb7"
time="2020-11-26T08:31:12Z" level=info msg="Node ID: kurediya-worker2"
time="2020-11-26T08:31:12Z" level=info msg="Lock Annotation: kube-system/kured:weave.works/kured-node-lock"
time="2020-11-26T08:31:12Z" level=info msg="Lock TTL not set, lock will remain until being released"
time="2020-11-26T08:31:12Z" level=info msg="Reboot Sentinel: /var/run/reboot-required every 30s"
time="2020-11-26T08:31:12Z" level=info msg="Blocking Pod Selectors: []"
time="2020-11-26T08:31:12Z" level=info msg="Reboot on: SunMonTueWedThuFriSat between 00:00 and 23:59 UTC"
time="2020-11-26T08:31:41Z" level=info msg="Reboot not required"
```

Triggering on one of the nodes reboot:
(on slack)
```
kured-botAPP  9:32 AM
Draining node kurediya-worker
9:32
Rebooting node kurediya-worker
```

evrardjp (approved):
LGTM we can improve later :)

evrardjp:
One of the possible improvements is to extract the shoutrrr.Send methods and expose them into some kind of Notify functions, for which we can build proper tests. 

Another one is to remove the previous slack notification code. For this, we can (where we parse the CLI arguments for example) format a notifyURL based on the previous slack URL arguments. Then everything can be given into shoutrrr.

dholbach:
Pushed `docker.io/weaveworks/kured:atighineanu-master-da19826` for testing.

dholbach:
I had a brief look at the shoutrrr API and it seems to allow sending to more than one URL. It this worth supporting in kured too, now that we're introducing a new field anyway?

atighineanu:
> I had a brief look at the shoutrrr API and it seems to allow sending to more than one URL. It this worth supporting in kured too, now that we're introducing a new field anyway?

If it's needed, we can introduce the functionality later. I didn't test that (the multiple URL thing) extensively (yet), because the main issue was to make possible to notify not only in slack. I'd need to test it to make sure there won't be any unexpected segfaults.

evrardjp:
> I had a brief look at the shoutrrr API and it seems to allow sending to more than one URL. It this worth supporting in kured too, now that we're introducing a new field anyway?

Yes it was discussed in the other PR  with @DaniJG, when we discussed the message of the notifications to be sent.

I agree this should be done, if possible in another PR. Sorry I forgot to mention in my later improvements.

dholbach:
> I agree this should be done, if possible in another PR. Sorry I forgot to mention in my later improvements.

How would this look like in cli arguments or helm values? Would we need to rename any of the args afterwards to support more than one URL?

evrardjp:
> > I agree this should be done, if possible in another PR. Sorry I forgot to mention in my later improvements.
> 
> How would this look like in cli arguments or helm values? Would we need to rename any of the args afterwards to support more than one URL?

Ahah you're too fast! I just edited my comment above :)

dholbach:
From our Dev environment:

Message from a few days ago (before this PR):
Screenshot from 2020-11-27 11-14-31

With `docker.io/weaveworks/kured:atighineanu-master-da19826`:
Screenshot from 2020-11-27 11-26-30

This was with no other config changes, so expectedly this turned up in our logs as well:
```
2020-11-27T10:09:02.128476036Z time="2020-11-27T10:09:02Z" level=warning msg="Deprecated flag(s). Please use --notify-url flag instead."
```

dholbach:
This will need an addition in the helm chart too, right?

evrardjp:
Agreed @dholbach

evrardjp:
Also the code doesn't seem right for my gofmt. Is that only me?

cmd/kured/main.go
```diff
@@ -42,6 +43,7 @@ var (
 	slackHookURL           string
 	slackUsername          string
 	slackChannel           string
+	notifyURL	           string
```
evrardjp, line 46: is this correctly aligned?

cmd/kured/main.go
```diff
@@ -114,6 +119,16 @@ func main() {
 	}
 }
 
+func flagcheck(cmd *cobra.Command, args []string) {
```
evrardjp, line 122: generally the name of the functions are following some kind of Camel case, so I would expect this function to be named flagCheck or flagChecker

cmd/kured/main.go
```diff
@@ -67,6 +69,7 @@ func main() {
 	rootCmd := &cobra.Command{
 		Use:   "kured",
 		Short: "Kubernetes Reboot Daemon",
+		PreRun: flagcheck,
```
evrardjp, line 72: I am not familiar with cobra, so I trust more experimented folks to determine if PreRun is the right place to evaluate and validate the CLI arguments.

charts/kured/templates/daemonset.yaml
```diff
@@ -81,6 +81,9 @@ spec:
           {{- if .Values.configuration.slackUsername }}
             - --slack-username={{ .Values.configuration.slackUsername }}
           {{- end }}
+          {{- if .Values.configuration.notifyURL }}
+            - --notify-url={{ .Values.configuration.notifyURL }}
+          {{- end }}
           {{- if .Values.configuration.messageTemplateDrain }}
```
evrardjp, line 87: If we change the helm chart, we should probably bump its version.

evrardjp (changes requested):
A few extra nits.

charts/kured/templates/daemonset.yaml
dholbach: AFAICS it's already bumped from another PR, but not released yet, so we should be good.

evrardjp: Oh that's interesting. I am not sure how chart-testing action works on the linting, does it bases itself on what has been already released, or on the existing code? If it's existing code, it means we can end up skipping releases.

@ckotzbauer do you think it's okay to do it this way, or should we edit the chart linting action? For me it kinda makes sense to bump the chart at the same time as we modify code, and release when we are ready, even if it means skipping a release.

cmd/kured/main.go
chentex: I feel it is a good place since it will not be inherited to the children, and the arguments should be loaded also.

dholbach:
Pushed `docker.io/weaveworks/kured:atighineanu-master-d92ac65` for testing.

dholbach:
Screenshot from 2020-12-08 10-24-19

Was there supposed to be a notification about draining as well?

dholbach:
Maybe we could file an follow-up issue for notifying more than one URL?

evrardjp:
@atighineanu can you fixup 9244778ac6358c7d9d5307a01e6c797245504e5a ? The title is still wrong, and the author is incorrect :p

evrardjp:
> Maybe we could file an follow-up issue for notifying more than one URL?

Sure let me do that.

dholbach:
The only open question I have before merging is
> Was there supposed to be a notification about draining as well?

evrardjp:
> The only open question I have before merging is
> 
> > Was there supposed to be a notification about draining as well?

Yeah it's not the first time this question was raised :) 
I suppose draining notification works without this patch? 

I don't notify at all in my test clusters, making it hard for me to test it.

atighineanu:
> @atighineanu can you fixup [9244778](https://github.com/weaveworks/kured/commit/9244778ac6358c7d9d5307a01e6c797245504e5a) ? The title is still wrong, and the author is incorrect :p

Done. Thanks.

atighineanu:
> > The only open question I have before merging is
> > > Was there supposed to be a notification about draining as well?
> 
> Yeah it's not the first time this question was raised :)
> I suppose draining notification works without this patch?
> 
> I don't notify at all in my test clusters, making it hard for me to test it.

Strange. I've repeatedly checked it with kind and I got both notifications on slack and rocketchat.

dholbach:
Maybe also add a commented-out option to `kured-ds.yaml`?

dholbach:
`docker.io/weaveworks/kured:atighineanu-master-da7f425`

dholbach:
I get a segfault when trying to contact mattermost - I probably set it up the wrong way, but maybe we need to catch an error somewhere?

```
time="2020-12-08T13:29:48Z" level=warning msg="Error notifying: failed to send notification to service, response status code 404 Not Found"
WARNING: ignoring DaemonSet-managed Pods: kube-system/kube-proxy-8wg77, kube-system/kured-t2mql; deleting Pods not managed by ReplicationController, ReplicaSet, Job, DaemonSet or StatefulSet: kube-system/storage-provisioner
evicting pod kube-system/storage-provisioner
evicting pod kube-system/coredns-f9fd979d6-9g8x4
time="2020-12-08T13:30:00Z" level=info msg="Commanding reboot for node: minikube"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x133fc8c]

goroutine 9 [running]:
github.com/containrrr/shoutrrr/pkg/services/mattermost.(*Service).Send(0xc0009f5480, 0xc000a49540, 0x17, 0xc000611548, 0xc0009f5480, 0x0)
	/home/daniel/go/pkg/mod/github.com/containrrr/shoutrrr@v0.0.0-20201124203606-785738593562/pkg/services/mattermost/mattermost.go:38 +0x20c
github.com/containrrr/shoutrrr.Send(0x7fff3e673b9e, 0x44, 0xc000a49540, 0x17, 0x1, 0xc000a49540)
	/home/daniel/go/pkg/mod/github.com/containrrr/shoutrrr@v0.0.0-20201124203606-785738593562/shoutrrr.go:24 +0xcd
main.commandReboot(0xc00003e02e, 0x8)
	/home/daniel/dev/kured/cmd/kured/main.go:313 +0x280
main.rebootAsRequired(0xc00003e02e, 0x8, 0xc0000b0f80, 0x0)
	/home/daniel/dev/kured/cmd/kured/main.go:380 +0x523
created by main.root
	/home/daniel/dev/kured/cmd/kured/main.go:414 +0x637
```
This is with da7f425.

evrardjp:
Here are my unsollicited 2cents:

I would restrict this PR to just implementing a generic mechanism, and check for parity with existing (=slack).
 
So for me, a shoutrrr issue with mattermost should be handled separately, within shoutrrr, not here (the error message is quite explicit and can be reproduced).

I think in the future we should implement tests, either in shoutrrr or in kured, and define the expectations within kured (should we stop if we fail to notify, or not?).

However, I agree that the current error message doesn't look good :)

README.md
```diff
@@ -225,6 +225,22 @@ Alternatively you can use the `--message-template-drain` and `--message-template
...
+
+When you specify a formatted URL using `--notify-url`, kured will notify
+about draining and rebooting nodes across a list of technologies. Here is 
+the syntax:
+
+slack:           `slack://tokenA/tokenB/tokenC`
+
+rocketchat:      `rocketchat://[username@]rocketchat-host/token[/channel|@recipient]`
+
+teams:           `teams://token-a/token-b/token-c`
```
evrardjp, line 238: have all of those been tested and working?

atighineanu: RC was tested. No idea about teams. You can check the performance by the number of the issues there.

atighineanu:
> Here are my unsollicited 2cents:
> 
> I would restrict this PR to just implementing a generic mechanism, and check for parity with existing (=slack).
> 
> So for me, a shoutrrr issue with mattermost should be handled separately, within shoutrrr, not here (the error message is quite explicit and can be reproduced).
> 
> I think in the future we should implement tests, either in shoutrrr and/or in kured, and define the expectations within kured (should we stop if we fail to notify, or not?).
> 
> However, I agree that the current error message doesn't look good :)

Indeed it is worth discussing if it should crashloopbackoff because of a failed notification or not. It whether could be triggered  manually with a flag or can simply send only Warnings. What you guys think?

README.md
evrardjp: If it's not tested, I would remove it. People can extend the documentation when we have tested and ensure working.

tonedefdev:
Any updates on when this PR will get merged and released? I'm really looking forward to being able to use Teams for notifications

dholbach:
Currently it needs to be rebased again. Not quite sure if there are any issues right now - I'd be happy to test it again.

dholbach:
Superseded by #315.

PR #237, opened 2020-11-23
Drain: allow pods grace period to terminate

### LINKED ISSUE #236: Drain doesn't wait for pods to terminate
This is an unfortunate consequence of #210; the default behaviour of `drain.Helper` is to delete immediately.

@evrardjp spotted that adding a line like this should fix it: https://github.com/kubernetes/kubernetes/blob/540e41c825f7/staging/src/k8s.io/kubectl/pkg/cmd/drain/drain.go#L148

bboreham:
The default of 0 is taken as "delete immediately", which is not appropriate.

Fixes #236

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -244,6 +244,7 @@ func drain(client *kubernetes.Clientset, node *v1.Node) {
 
 	drainer := &kubectldrain.Helper{
 		Client:              client,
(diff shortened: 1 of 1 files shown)


dholbach (approved):
Thanks a lot Bryan for figuring this out with Jean-Philippe and documenting this in k/k as well!

PR #238, opened 2020-11-24
rename annotation-ttl to lock-ttl in all places, follow-up to #213

dholbach:
The command-line option was changed in #213 and needed to be updated in the chart and docs.


dholbach:
CC @mvisonneau

ckotzbauer (approved):

PR #246, opened 2020-11-27
Improve coordinated reboot output

evrardjp:
When a failure is happening and the cluster doesn't manage to
be back up on time, we exit 1, and don't show docker logs.

This is a problem, as we would benefit from a detailed docker
output on those cases, when debugging.

This fixes it by ensuring the logging is always done at the
exit of the script.


ckotzbauer (approved):

dholbach (approved):

evrardjp:
Oopsies, I didn't put WIP there. I didn't have the full time to test it. Let's see if I have broken stuff then!

PR #248, opened 2020-11-27
Add Shellcheck

evrardjp:
Ensures our bash is neat!


evrardjp:
That won't pass until #247 is merged.

dholbach (approved):

Issue #19, opened 2018-05-07
taint node with PreferNoSchedule until it gets drained


damoon:
In case of multiple nodes asking to restart, pods will be moved more often then needed, because they can start on the node that will be rebooted next.
The use of PreferNoSchedule on nodes waiting to be rebooted, will prefer scheduling pods onto already rebooted nodes.

link: https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/

damoon:
@weaveworksbot i am still interested in this

bboreham:
Isn't there a danger that every node is waiting to reboot (e.g. if Ubuntu push out a security patch), so this would mean no pods can get scheduled?

damoon:
The taint is PreferNoSchedule not NoSchedule.
The scheduler will try to avoid putting pods on these nodes, but do it if no other placement is possible.

PR #250, opened 2020-11-28
implement issue-19 add prefer no schedule taint to avoid double draining of pods

damoon:
This implements https://github.com/weaveworks/kured/issues/19

The idea is to set a prefer no schedule taint.
The taint is set when the reboot sentinel is found and removed after the restart.

In case the node can not restart immediately the taint asks the kubernetes scheduler to avoid placing new pods on the tainted node.
But this is only preferred, as a last resort the kubernetes scheduler does place pods onto the node.

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -87,8 +87,8 @@ func main() {
 		"alert names to ignore when checking for active alerts")
 	rootCmd.PersistentFlags().StringVar(&rebootSentinel, "reboot-sentinel", "/var/run/reboot-required",
 		"path to file whose existence signals need to reboot")
-	rootCmd.PersistentFlags().StringVar(&preferNoScheduleTaintName, "prefer-no-schedule-taint", "weave.works/kured-node-reboot",
-		"Taint name applied during pending node reboot (to prevent receiving additional pods from other rebooting nodes). Set to \"\" to disable tainting.")
+	rootCmd.PersistentFlags().StringVar(&preferNoScheduleTaintName, "prefer-no-schedule-taint", "",
+		"Taint name applied during pending node reboot (to prevent receiving additional pods from other rebooting nodes). Disabled by default. Set e.g. to \"weave.works/kured-node-reboot\" to enable tainting.")
 
 	rootCmd.PersistentFlags().StringVar(&slackHookURL, "slack-hook-url", "",
 		"slack hook URL for reboot notfications")
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/weaveworks/kured/pkg/daemonsetlock"
 	"github.com/weaveworks/kured/pkg/delaytick"
 	"github.com/weaveworks/kured/pkg/notifications/slack"
+	"github.com/weaveworks/kured/pkg/taints"
 	"github.com/weaveworks/kured/pkg/timewindow"
 )
 
@@ -31,20 +32,21 @@ var (
 	version = "unreleased"
 
 	// Command line flags
-	period                time.Duration
-	dsNamespace           string
-	dsName                string
-	lockAnnotation        string
-	lockTTL               time.Duration
-	prometheusURL         string
-	alertFilter           *regexp.Regexp
-	rebootSentinel        string
-	slackHookURL          string
-	slackUsername         string
-	slackChannel          string
-	messageTemplateDrain  string
-	messageTemplateReboot string
-	podSelectors          []string
+	period                    time.Duration
+	dsNamespace               string
+	dsName                    string
+	lockAnnotation            string
+	lockTTL                   time.Duration
+	prometheusURL             string
+	alertFilter               *regexp.Regexp
+	rebootSentinel            string
+	preferNoScheduleTaintName string
+	slackHookURL              string
+	slackUsername             string
+	slackChannel              string
+	messageTemplateDrain      string
+	messageTemplateReboot     string
+	podSelectors              []string
 
 	rebootDays  []string
 	rebootStart string
@@ -85,6 +87,8 @@ func main() {
 		"alert names to ignore when checking for active alerts")
 	rootCmd.PersistentFlags().StringVar(&rebootSentinel, "reboot-sentinel", "/var/run/reboot-required",
 		"path to file whose existence signals need to reboot")
+	rootCmd.PersistentFlags().StringVar(&preferNoScheduleTaintName, "prefer-no-schedule-taint", "weave.works/kured-node-reboot",
+		"taint name applied during pending node reboot (to prevent receiving additional pods from other rebooting nodes)")
 
 	rootCmd.PersistentFlags().StringVar(&slackHookURL, "slack-hook-url", "",
 		"slack hook URL for reboot notfications")
@@ -336,10 +340,19 @@ func rebootAsRequired(nodeID string, window *timewindow.TimeWindow, TTL time.Dur
 		release(lock)
 	}
 
+	preferNoScheduleTaint := taints.New(client, nodeID, preferNoScheduleTaintName, v1.TaintEffectPreferNoSchedule)
+
+	// Remove taint immediately during startup to quickly allow scheduling again.
+	if !rebootRequired() {
+		preferNoScheduleTaint.Disable()
+	}
+
 	source := rand.NewSource(time.Now().UnixNano())
 	tick := delaytick.New(source, period)
 	for range tick {
 		if !window.Contains(time.Now()) {
+			// Remove taint outside the reboot time window to allow for normal operation.
+			preferNoScheduleTaint.Disable()
 			continue
 		}
 
@@ -358,6 +371,8 @@ func rebootAsRequired(nodeID string, window *timewindow.TimeWindow, TTL time.Dur
 		nodeMeta.Unschedulable = node.Spec.Unschedulable
 
 		if !acquire(lock, &nodeMeta, TTL) {
+			// Prefer to not schedule pods onto this node to avoid draing the same pod multiple times.
+			preferNoScheduleTaint.Enable()
 			continue
 		}
 
@@ -392,6 +407,7 @@ func root(cmd *cobra.Command, args []string) {
 	} else {
 		log.Info("Lock TTL not set, lock will remain until being released")
 	}
+	log.Infof("PreferNoSchedule taint: %s", preferNoScheduleTaintName)
 	log.Infof("Reboot Sentinel: %s every %v", rebootSentinel, period)
 	log.Infof("Blocking Pod Selectors: %v", podSelectors)
 	log.Infof("Reboot on: %v", window)
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -31,20 +31,20 @@ var (
 	version = "unreleased"
 
 	// Command line flags
-	period                 time.Duration
-	dsNamespace            string
-	dsName                 string
-	lockAnnotation         string
-	lockTTL                time.Duration
-	prometheusURL          string
-	alertFilter            *regexp.Regexp
-	rebootSentinel         string
-	slackHookURL           string
-	slackUsername          string
-	slackChannel           string
-	messageTemplateDrain   string
-	messageTemplateReboot  string
-	podSelectors           []string
+	period                time.Duration
+	dsNamespace           string
+	dsName                string
+	lockAnnotation        string
+	lockTTL               time.Duration
+	prometheusURL         string
+	alertFilter           *regexp.Regexp
+	rebootSentinel        string
+	slackHookURL          string
+	slackUsername         string
+	slackChannel          string
+	messageTemplateDrain  string
+	messageTemplateReboot string
+	podSelectors          []string
 
 	rebootDays  []string
 	rebootStart string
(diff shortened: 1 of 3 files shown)


damoon:
The test is failing because the taint code uses the update method, but RBAC does only include patching nodes.

damoon:
it works now :smiley: 
someone wants to review it?

cmd/kured/main.go
```diff
@@ -336,26 +436,44 @@ func rebootAsRequired(nodeID string, window *timewindow.TimeWindow, TTL time.Dur
...
+		}
+
+		node, err := client.CoreV1().Nodes().Get(context.TODO(), nodeID, metav1.GetOptions{})
+		if err != nil {
+			log.Fatal(err)
+		}
+		nodeMeta.Unschedulable = node.Spec.Unschedulable
+
+		if !acquire(lock, &nodeMeta, TTL) {
+			continue
```
evrardjp, line 467: I am confused as how preferNoSchedule works. 

I thought that the preferNoSchedule flag would be set (3rd argument to true) when:
- The rebootRequired returns true
- The node cannot get a lock because another node has a lock.
Therefore I expected the line 454 to be in here (L467) instead, before the continue. However, if we agree that all nodes should be marked as NoSchedule immediately when we know they need rebooting (during the maintenance window), then it's fine for me. After all, the code is indeed cleaner in the current PR implementation.
As long as we document how we do it (maybe a little comment on L454 would help?).

However, I expected the function to flip the NoSchedule to false at all ticks, not simply at the restart of the daemonset. If the timewindow to trigger all the reboots is too small, some nodes might not have rebooted in the maintenance window, and therefore some ndoes will still marked with no schedule outside the timewindow. That doesn't seem right.

TL;DR: Therefore I would move the L440 inside the block 450-452, move that block above the block of the timewindow (446), move the L454 `preferNoSchedule` in the block L466-468 before the `continue`. But of course, I am easy, and happy to discuss it here.

evrardjp (changes requested):
First, allow me to thank you for this PR! Well thought out.

Please see my comments inline for the first part of the review (maybe a comment, and a little code change?) on the preferNoSchedule code location/behaviour.
Finally, I see a problem with the PR right now (maybe I am mistaken, could you clarify?):
This PR changes the state of the taint `TaintEffectPreferNoSchedule` regardless of not it was tained by kured. This is a problem for me, because one administrator might want to mark a node for NoSchedule for other reasons, and we might remove it without its approval. WDYT?

cmd/kured/main.go
```diff
@@ -85,6 +88,8 @@ func main() {
 		"alert names to ignore when checking for active alerts")
 	rootCmd.PersistentFlags().StringVar(&rebootSentinel, "reboot-sentinel", "/var/run/reboot-required",
 		"path to file whose existence signals need to reboot")
+	rootCmd.PersistentFlags().StringVar(&preferNoScheduleTaint, "prefer-no-schedule-taint", "weave.works/kured-node-reboot",
+		"taint to avoid double drained pods")
```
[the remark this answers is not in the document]
damoon: i like it.

cmd/kured/main.go
damoon: i assumed the order of the checks would be like this because the coast to execute them.
the window-check is just a calculation, the requires-reboot is a local check and the reboot-blocked check is a rpc.

i did not want to have the remove-taint within the loop, because it could take a long time until a rebooted node get untained.
maybe a additional untaint in the block from 450-453?

The true-false flag was a pain in my eye anyway. What do you think about two functions addPreferNoScheduleTaint and removePreferNoScheduleTaint?

damoon: > This PR changes the state of the taint `TaintEffectPreferNoSchedule` regardless of not it was tained by kured. This is a problem for me, because one administrator might want to mark a node for NoSchedule for other reasons, and we might remove it without its approval. WDYT?

Only the taint "weave.works/kured-node-reboot" (by default) is added or remove by the new behavior. Other taints will be untouched. The code from https://github.com/damoon/kured/blob/19-PreferNoSchedule/cmd/kured/main.go#L326 until https://github.com/damoon/kured/blob/19-PreferNoSchedule/cmd/kured/main.go#L367 ensures only to add/remove this one taint.

damoon:
> This PR changes the state of the taint TaintEffectPreferNoSchedule regardless of not it was tained by kured. This is a problem for me, because one administrator might want to mark a node for NoSchedule for other reasons, and we might remove it without its approval. WDYT?

Only the taint "weave.works/kured-node-reboot" (by default) is added or remove by the new behavior. Other taints will be untouched. The code from https://github.com/damoon/kured/blob/19-PreferNoSchedule/cmd/kured/main.go#L326 until https://github.com/damoon/kured/blob/19-PreferNoSchedule/cmd/kured/main.go#L367 ensures only to add/remove this one taint.

cmd/kured/main.go
evrardjp: For addPreferNoScheduleTaint: That's indeed more explicit (especially in this part of the code). But overall it's up to you, you wrote it. It's no big deal for me. However: Let's not copy the function into two (and repeat ourselves):  that sounds even worse :)

I am not sure it's about the cost to execute. Someone with more history of the project shoudl tell :)

I was not fond of having this additional untaint without explanations. For me, all this needs is comments, to explain what the design decisions we made. For example:
`we do this outside the main loop because it's an API task, so it takes more time, for which we want to avoid running in the tick loop. This means that a temporary API issue will cause a crash of the daemonset until its state is correct. This also means that if the maintenance window is too small, the machines can still technically be marked with no schedule outside out the maintenance window`.

damoon:
I was not sure who the code was organized.
Should this get its own package in https://github.com/weaveworks/kured/tree/master/pkg ?

cmd/kured/main.go
damoon: I added a comment for the untaint outside the loop,
moved the tainting inside the check for rebootBlocked,
and added a untaint to allow for the the reboot window.

Taints got their own package and better naming, the bool is hidden.

Does this look good to you?

evrardjp (commented):
I have not tested it, but I like it.

cmd/kured/main.go
evrardjp: :+1:

evrardjp:
CI tested it, so I guess that's a good start.

Should we extend smoke test to ensure that we can schedule a basic workload?

damoon:
I tested the parallel reboot of nodes and the taint creation/removal.
(But this happened before moving the code to its own package.)
I would feel more comfortable, if someone else would check it too.

damoon:
i found an issue:
rebootBlocked checks prometheus for alerts and does not indicate an other node has the lock
i moved adding the taint into the !acquire block.

once the testing is done is push the fix.

damoon:
fixed and tested. works for me.

pkg/taints/taints.go
```diff
@@ -0,0 +1,140 @@
...
+				Op:    "add",
+				Path:  "/spec/taints/-",
+				Value: taint,
+			},
+		}
+	}
+
+	patchBytes, err := json.Marshal(patches)
+	if err != nil {
+		log.Fatalf("Error encoding taint patcht for node %s: %v", nodeID, err)
```
bboreham, line 127: nit: patcht typo

pkg/taints/taints.go
```diff
@@ -0,0 +1,140 @@
...
+func (t *Taint) Enable() {
+	preferNoSchedule(t.client, t.nodeID, t.taintName, t.effect, true)
+}
+
+// Disable removes the taint for a node. Removing a missing taint is a noop.
+func (t *Taint) Disable() {
+	preferNoSchedule(t.client, t.nodeID, t.taintName, t.effect, false)
+}
+
+func preferNoSchedule(client *kubernetes.Clientset, nodeID, taintName string, effect v1.TaintEffect, taintShouldExists bool) {
```
bboreham, line 43: `ShouldExist` ?

cmd/kured/main.go
```diff
@@ -336,10 +340,19 @@ func rebootAsRequired(nodeID string, window *timewindow.TimeWindow, TTL time.Dur
...
+	if !rebootRequired() {
+		preferNoScheduleTaint.Disable()
+	}
+
 	source := rand.NewSource(time.Now().UnixNano())
 	tick := delaytick.New(source, period)
 	for range tick {
 		if !window.Contains(time.Now()) {
+			// Remove taint outside the reboot time window to allow for normal operation.
+			preferNoScheduleTaint.Disable()
```
bboreham, line 355: I'm not entirely happy that this is hitting the API server every time round the loop, when most of the time it will have nothing to do.  

Maybe hold a flag saying whether the taint was previously added?

cmd/kured/main.go
```diff
@@ -85,6 +87,8 @@ func main() {
 		"alert names to ignore when checking for active alerts")
 	rootCmd.PersistentFlags().StringVar(&rebootSentinel, "reboot-sentinel", "/var/run/reboot-required",
 		"path to file whose existence signals need to reboot")
+	rootCmd.PersistentFlags().StringVar(&preferNoScheduleTaintName, "prefer-no-schedule-taint", "weave.works/kured-node-reboot",
```
bboreham, line 90: Maybe allow `""` as an option to disable the feature?

bboreham (commented):
I left some minor comments below.

It would be good to mention this feature in the README.

I did not really follow the `test` / `add` patch stuff.  Do you have a reference to some documentation explaining it, or another program that does the same?  (`kubectl taint` does not appear to do it).

Also I think there is another corner case: if `rebootRequired()` goes from true to false the taint will not be removed until the end of the reboot window, or until restart of kured.

pkg/taints/taints.go
damoon: renamed :D

pkg/taints/taints.go
damoon: fixed type

cmd/kured/main.go
damoon: done

cmd/kured/main.go
damoon: done

damoon:
I updated the parameter section in README.

For patches I found this https://dwmkerr.com/patching-kubernetes-resources-in-golang/ example and for `test` i found http://jsonpatch.com/#test. \
Test is needed to ensure the list of taints did not change on server side while the offset to edit was processed locally. \
I believe kubectl uses update instead of patch and avoids this problem with a optimistic lock. \
This implementation chose patch because patch was already part of the RBAC profile while update node is missing.

I changed the behavior in case `rebootRequired()` switches from true to false to remove the taint.

bboreham (approved):
Looks good to me.
Thanks for all changes, and for the explanation of why it is done this way.

cmd/kured/main.go
```diff
@@ -336,26 +340,50 @@ func rebootAsRequired(nodeID string, window *timewindow.TimeWindow, TTL time.Dur
...
+			continue
+		}
+
+		if !rebootRequired() {
+			preferNoScheduleTaint.Disable()
+			continue
+		}
+
+		if rebootBlocked(client, nodeID) {
+			continue
```
evrardjp, line 365: if the reboot is Blocked by a pod or prometheus, shouldn't we still unmark the node?
For me it means that this node was voluntarily put out of the loop by an administrator, even if the reboot is required during the maintenance window.

In this case, I think that rebootBlocked should also remove any existing NoSchedule taint.
On the other hand, one could argue that a node "out of the loop" shouldn't be touched _at all_. 

In any case, this decision should be recorded somewhere (in a comment, or in the commit message).

evrardjp (commented):
I have added a question/comment to clarify our goals. I would prefer if this is clarified before merging. However, I don't think it's worth blocking this PR from merging if we don't get consensus.

Please note that I didn't test this patch. The code looks okay to me.

cmd/kured/main.go
bboreham: Being blocked by prometheus can be from a "pending" alert: in my case I see those come and go quite often.
So I think it would be bad to oscillate the taint; I would say once you need to reboot you shouldn't be scheduling new pods onto that node.  (But it's only "preferred" so may get some if necessary).

Maybe we should make the taint `""` by default, so people can choose to try out the feature and report back?

evrardjp: I am fine with both parts of your answer.

evrardjp: @damoon do you mind updating the taint to "" by default?

evrardjp: Alternatively, we can merge this after 16.0, and I can do another PR with updating the default taint.

damoon: I changed the flag to be disabled by default.

damoon: Thinking about the rebootBlocked() cases.

Situation: A node that waits for a long time to reboot, but is blocked by other rebooting nodes. Then a pod, that blocks reboots, is scheduled onto that node. 
A) The taint is remove. As a result even more work gets scheduled here, and needs to be drained again. This should be avoided.
B) The taint stays. As a result the nodes without the taint get a higher average load, even so this node can not reboot once it is its turn to do so. This should also be avoided.
The same should be true for prometheus alert.

My feeling is, to leave a node, that wants to reboot, but is blocked to do so, tainted. It expresses the intend best.
As mentioned, the taint is "only" preferred, and in case B i trust the kubernetes scheduler and resource requests enough for uneven load not be a real issue.

evrardjp: I agree.

bboreham (approved):
lgtm, thanks!

dholbach:
Nice work and thanks for hanging in there.

Do you think you could add a note https://github.com/weaveworks/kured/issues/295 about this feature?

PR #258, opened 2020-11-30
Publish image on tag with github actions

evrardjp:
As we are pretty much committed to github actions, we should
probably rely on it to push the images on tag.

This covers the missing bits.


ckotzbauer (approved):

dholbach:
I pinged out Corp IT folks to take a look at this.

What can we do to test this?

evrardjp:
I would say that the easiest is to start using it, and start tagging :)

dholbach:
Ok. I have the necessary steps from our Corp IT folks. I guess I can start taking a look at Friday, but this is not necessarily urgent, right - we need to get mainline closer to be in a releasable state first?

evrardjp:
correct.

README.md
```diff
@@ -283,8 +283,7 @@ Using `--lock-ttl=30m` will allow other nodes to take over if TTL has expired (i
 
 ## Building
 
-See the [CircleCI config](.circleci/config.yml) for the preferred
-version of Golang. Kured now uses [Go
+Kured now uses [Go
```
dholbach, line 286: Maybe point to `go.mod` instead now?

dholbach:
Thanks a lot for the fix @morancj and the work behind the scenes!

@evrardjp Does this generally look good to go to you?

README.md
evrardjp: Yes, I could do that in a follow up patch, or in this one.

evrardjp (commented):
I can't judge on the secret names, so I trust you folks :)

evrardjp:
I have updated this patch to point to golang version.

dholbach:
Hum, why did no GH actions run for the last commit? https://github.com/weaveworks/kured/pull/258/checks

dholbach:
Got it, just saw https://github.com/weaveworks/kured/pull/262

evrardjp:
> Got it, just saw #262

My bad!

morancj:
> The advantage of using the names I set was that it was not too weird when ppl fork the repository, tag, and push on their docker hub. But I think it's fine.

@evrardjp That would be nice, but I don't think we can do this for Org. level secrets (which `DOCKERHUB_TOKEN_WEAVEWORKSKUREDCI` is). We could use [Encrypted secrets](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#limits-for-secrets), but that would still need renaming on a fork, is more work, and not much easier to figure out for anyone who forks. Happy to be corrected if you know of a mechanism or have a better suggestion!

evrardjp:
I think it's fine, and we shouldn't worry too much about this. The idea of this PR is to make the whole thing more transparent, and I think it achieves it right now. If people carry a fork, then they can probably add a patch for their org or rename the secret. LGTM as it is now.

morancj:
@evrardjp Agree, thanks. If it becomes an issue, we can investigate some hack like this at a later date: https://github.community/t/make-secrets-available-to-builds-of-forks/16166/21

evrardjp:
To clarify something: I don't promote that weaveworks secret are made available to forks, quite the contrary. We want each fork have its own set of credentials. Only the main repo (weaveworks/kured) needs to work with weavework's credentials.

dholbach:
Can you rebase?

Does this also do builds of each commit? We currently have them on Dockerhub too: https://hub.docker.com/r/weaveworks/kured/tags?page=1&ordering=last_updated

evrardjp:
@dholbach rebased.

evrardjp:
This does NOT build on every merge, but we could do it, in a follow up patch, if you like. It will be trivial.

dholbach:
I think it'd be good to include this in this PR, so we have images for all commits and can see if things work all right?

evrardjp:
I have another PR opened, let me close it quick and add its content here.

evrardjp:
done.

dholbach (approved):
Super, thanks for looking into this.

PR #259, opened 2020-11-30
Force golang version

evrardjp:
Without this, golang version used is the golang version decided
by github.

This is a problem, as it might shift over time, without our control.

This fixes it by getting the golang version from the go.mod.

### DIFF
--- a/.github/workflows/on-release.yaml
+++ b/.github/workflows/on-release.yaml
@@ -9,9 +9,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-        with:
-          fetch-depth: "0"
-
       - name: Publish Helm chart
         uses: stefanprodan/helm-gh-pages@master
         with:
--- a/.github/workflows/on-pr.yaml
+++ b/.github/workflows/on-pr.yaml
@@ -17,6 +17,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    - name: Find go version
+      run: |
+        GO_VERSION=$(awk '/^go/ {print $2};' go.mod)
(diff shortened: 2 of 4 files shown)


.github/workflows/on-pr-charts.yaml
```diff
@@ -22,8 +22,6 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-        with:
-          fetch-depth: "0"
```
ckotzbauer, line 26: I'm not sure if this can be removed. The chart-testing-action uses the changes to find the charts which are changed. That's the default behavior unless it is specified otherwise in the ct.yaml

evrardjp: good point! I removed it everywhere, but that won't work in here.

ckotzbauer (approved):

.github/workflows/on-release.yaml
```diff
@@ -9,9 +9,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-        with:
-          fetch-depth: "0"
-
```
dholbach, line 14: This was added recently IIRC - what was is it about?

evrardjp: I removed it everywhere it's not necessary.

Instead of doing a shallow clone, it does a git clone and fetches the whole history.

evrardjp: If necessary, I can extract this into another PR, but I have the impression we have more PRs than merges, so I didn't want to spam PRs, like I usually do.

dholbach: :+1:

dholbach (approved):

PR #262, opened 2020-12-07
Fix typo in github workflows

evrardjp:
Without this patch, the PR jobs are broken and no jobs are running.
This was a recently introduced typo in the last refactor of the
PR jobs.

This should fix it, and make the PR test working again.


dholbach (approved):
LGTM

PR #263, opened 2020-12-07
Publish image on master merged changes

evrardjp:
As we are pretty much committed to github actions, we should
probably rely on it to push the images at each commit merged
on the master branch.


evrardjp:
Superseded by #258

PR #265, opened 2020-12-07
Remove circle ci configuration from gh pages

evrardjp:
We removed Circle CI from this repo, as we can do everything
from github actions. There is no point in keeping this
configuration here. Removing.


dholbach (approved):
LGTM

Issue #267, opened 2020-12-09
Test again k8s 1.{18,19,20}.x


dholbach:
In prep for the 1.6.0 release, we should test against the latest 3 releases now that k8s 1.20.0 is out.

evrardjp:
I will tacke the update of the gh actions.

evrardjp:
The CI will only test so much, so I suppose we need manual testing of prometheus/slack before closing this issue.

dholbach:
:heavy_check_mark: Smoketest against local minikube 1.20.1, 1.19.6, 1.18.14 with 86fe6ff (no prometheus, slack testing).

dholbach:
Looks like things are working all right in our Dev cluster too, Slack notifications turned up fine as well.

dholbach:
Suggested release notes:
-------------------------------------

From a user perspective, this release will mostly bring you up to date with Kubernetes 1.[18,19,20].x. Behind the scenes Jean-Philippe Evrard (SuSE) put a lot of effort into automated testing, which will give us more confidence in the code we merge.

# Improvements
- add additional parameters to override the drain/reboot slack messages
- rename message template parameters so they are not related to slack
- Improve coordinated reboot output
- Add more logs into gates
- Added support for time wrap in timewindow.Contains


# Build / testing

- Bump to golang 1.15
- Add smoke/basic functional test
- Auto expire issues and PRs
- Always have latest helm binary installed
- Increase kubernetes versions test matrix for smoke tests
- Add kustomize testing
- Simplify manifest testing
- Add Shellcheck
- Cleanup github actions
- Fix chart linter
- Simplify action code
- Do not bump kubernetes with dependabot
- Force golang version
- Publish image on tag
- Point docs to current golang version
- Publish image on master merged changes
- Update for kubernetes 1.20 support
- Fix comment spacing
- Auto-publish helm chart on master change
- Do not bump any k8s module
- Temporarily workaround alpine issue

# Docs
- Clarify development process
- Clarify development process for helm charts
- Corrected README Setting a schedule configuration example

# Kubernetes Version Compatibility

The daemon image contains a 1.18.x k8s.io/{client-go,kubectl} for the purposes of maintaining the lock and draining worker nodes. Kubernetes aims to provide forwards & backwards compatibility of one minor version between client and server, so -this should work on 1.1{18,19,20}.x.

[...]

PR #268, opened 2020-12-09
Update for kubernetes 1.20 support

evrardjp:
This ensures we bump the code for 1.20.
It updates the testing to ensure kured works on a 1.20 cluster,
removes the testing on 1.17 (as it is now deprecated).
Libraries remain on 1.19, to avoid breaking 1.18 clusters.

Related: #267 .


charts/kured/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.5.1"
 description: A Helm chart for kured
 name: kured
-version: 2.2.2
+version: 2.2.3
```
dholbach, line 5: 2.2.2 is not released yet, so this is not necessary AFAICS: https://github.com/weaveworks/kured/tree/gh-pages

charts/kured/templates/clusterrole.yaml
```diff
@@ -9,9 +9,9 @@ rules:
 # Allow kured to read spec.unschedulable
 # Allow kubectl to drain/uncordon
 #
-# NB: These permissions are tightly coupled to the bundled version of kubectl; the ones below
-# match https://github.com/kubernetes/kubernetes/blob/v1.19.4/staging/src/k8s.io/kubectl/pkg/cmd/drain/drain.go
```
dholbach, line 13: Is the link useful for figuring out which apigroups, etc? If yes, we should probably keep it around?

evrardjp: Isn't the other text good enough? I am reluctant to keep it to a single version, but I think I can live with it.

evrardjp: Ok let me rollback that change, and keep it as is.

charts/kured/Chart.yaml
evrardjp: does it matter? the code with a helm chart of 2.2.2 was effectively out in the wild, even though not published through our gh pages...

Which reminds me that we should probably publish the helm chart on merge in master?

charts/kured/templates/clusterrole.yaml
evrardjp: Done.

charts/kured/Chart.yaml
dholbach: No, it needs a tag to publish.

cc @ckotzbauer

evrardjp: I have a PR to fix this: https://github.com/weaveworks/kured/pull/269

ckotzbauer: I reviewed #269. That's a good change for faster chart releases :+1:

dholbach (approved):

PR #269, opened 2020-12-09
Auto-publish helm chart on master change

evrardjp:
We are now testing the helm charts on each PR. They are now
ensured to be passing our tests and reviewed before merging.
This also means that the merged changes in the master branch
are reliable, and therefore can be consumed immediately.

Currently, we are waiting for a release to publish a helm
chart.

This is a problem as it means that the helm chart will
always lag behind, and we'll miss a few semantic versions,
if for example the helm chart is adapted multiple times
before the next release.

This should fix it by ensuring ALL the merged changes in
our helm chart will result in a new published helm chart.

### DIFF
--- a/.github/workflows/on-master-push-charts.yaml
+++ b/.github/workflows/on-master-push-charts.yaml
@@ -1,11 +1,14 @@
-name: Publish release
+name: Publish helm chart
 on:
-  release:
-    types: published
+  push:
+    branches:
+      - "master"
+    paths:
+      - "charts/**"
 
 jobs:
   publish-helm-chart:
-    name: Publish our chart
+    name: Publish latest chart
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -119,19 +119,6 @@ Then docker push the image. In the future, that might be automatically
 done when creating a tag on the repository, with the help of github
 actions.
 
-### Prepare Helm chart
-
-You should also check that the helm chart has been updated before issuing
-a release.  You can bump the helm chart with the latest image
-tag by running:
-
-```sh
-make DH_ORG="weaveworks" VERSION="1.3.0" helm-chart
-```
-
-Finally bump the `version` in `charts/kured/Chart.yaml` (following
-the versioning rules) and you should be all set.
-
 ### Create the combined manifest
 
 
@@ -149,10 +136,22 @@ cat kured-ds.yaml >> "$MANIFEST"
 Now you can head to the Github UI, use the version number as tag and upload the
 `kured-<release>-dockerhub.yaml` file.
 
-The creation of the release will publish the helm chart (thanks to github actions).
-
 Please describe what's new and noteworthy in the release notes, list the PRs
 that landed and give a shout-out to everyone who contributed.
 
 Please also note down on which releases the upcoming `kured` release was
 tested on. (Check old release notes if you're unsure.)
+
+### Update the Helm chart
+
+You can automatically bump the helm chart's application version
+with the latest image tag by running:
+
+```sh
+make DH_ORG="weaveworks" VERSION="1.3.0" helm-chart
+```
+
+A change in the helm chart requires a bump of the `version`
+in `charts/kured/Chart.yaml` (following the versioning rules).
+Update it, and issue a PR. Upon merge, that PR will automatically
+publish the chart to the gh-pages branch.


ckotzbauer (approved):

evrardjp:
Let's merge this _after_ 269.

dholbach:
#270 is landed - will you rebase?

Just so I understand this... will a chart be published for every commit?

evrardjp:
> #270 is landed - will you rebase?
> 
> Just so I understand this... will a chart be published for every commit?

Every commit merged on master, yes

dholbach:
>> Just so I understand this... will a chart be published for every commit?
>
> Every commit merged on master, yes

What will this mean for users? Usually we wait and do more testing before new releases get out.

evrardjp:
The new releases are thoroughly tested because they contain new code which is not covered by CI.

This would still be the case: When we publish a new "release" of kured code, we create a tag, then we create an image matching the tag. The helm chart is then updated, and we can update when we feel confident that the image is safe. In other words, anything updating the helm chart can be immediately pushed.

dholbach:
Sorry, I'm still not 100% clear on this. Currently I feel people are changing the Helm chart and the version all the time. Would that practice change then?

evrardjp:
We should behave the same way as if the helm chart was external: the helm chart is there to _consume_ the built images. Anything that changes the helm chart is welcome, but it needs to be done post release.

Some PRs include helm chart change, but that's technically incorrect. If we want to include a new feature in kured, we should have two PRs (like if it was external): One to add the feature in kured, one to add it in the helm chart after the new kured is published. Of course, that's only for user facing changes. Dependency updates and others shouldn't trigger 2 PRs, as the user interface is the same (values.yaml isn't changed)

evrardjp:
I have updated DEVELOPMENT documentation, for understanding.

dholbach:
@evrardjp @ckotzbauer I'm still a bit lost. The way `stefanprodan/helm-gh-pages` works is that it publishes charts for git tags. So if I understand things correctly we ask folks to

1. tag a kured release
2. bump the chart using "make <...>" to pick up the built image
3. then tag another release 

Isn't this more complicated?

ckotzbauer:
This PR would decouple the kured releases from the chart releases. The benefit from this is, that we can release "chart-only" changes immediately and folks don't have to wait for the next kured release, only because there was a bug in a yaml for example.

The drawback is, that we introduce two different release/tag-types which could be problematic for people or tools which are looking for new releases of kured through the releases page...

@evrardjp Please correct me if I'm wrong :wink:

evrardjp:
When you think about it, the kured code, official image, and helm chart were already decoupled.

When creating a release, we have to create an image. That image is created from a git sha reference, and for clarity, a git tag matching that reference. It's only when that image _becomes official_ (by being published) that the helm chart can consume it, and we make a change in the master branch to consume that image.

Here comes the drift problem:
Case1: We have to pre-emptively break the helm chart on master branch before release, by pointing to a non-existing image. If we accept that, it means we can't rely on the helm chart test, can't actively test the image in the master branch, and we can tag/create a release with a broken helm chart. I don't think it's a great idea
Case2: We tag an image with the _future_ release X.Y.Z before changing the helm chart code. We change the helm chart, which can now pass the test with image X.Y.Z. When all of this is merged, we are at code (new sha above X.Y.Z), and we create a release with version number X.Y.Z. From there, we either republish the image (and therefore can break, because we didn't test that new image) to match the image and the git tag, or we accept the drift that the image tagged X.Y.Z doesn't correspond to the active published release matching tag X.Y.Z. This doesn't sound great to me. 
Case3: We don't break the helm chart on master, and continue testing it. It means that code can go in master branch, and it will always get tested and deployed. If a change impacts the helm chart, it will automatically be published with the appropriate helm chart version. Publication of a release mean publication of the code of kured at that moment, which means that the image will be always at the right tag, and the manifest/helm chart consuming those might lag behind. All of that process is documented, and should be clearer.

TL;DR: As the helm chart version isn't the same as the AppVersion, I think it's fine to decouple those. What I meant above is that they already are decoupled. We tried to couple those in the past, and I don't believe it's a good idea.

You're right @ckotzbauer, "chart-only" changes can be consumed directly, automatically. I feel that this increase our test coverage _at all times_, by not needing to break the helm chart of master when releasing. Creating the helm chart continue to trail the releasing of the software, as usual.

Again, I think we had two different release cycles, we just tried to shove everything into one, but without the clarity/visibility of the testing we now have.

evrardjp:
@dholbach isn't the action itself `stefanprodan/helm-gh-pages` publishing the helm chart on github pages, simply, regardless of the trigger? That was my understanding.

In this case, everytime we do something with the helm chart, we'll use that action, and it will publish the helm chart. It doesn't need to _wait_ for the release for introducing a fix.

dholbach (approved):
thanks for explaining it again!

PR #275, opened 2020-12-14
Do not bump any k8s module

evrardjp:
Without this patch, dependabot will still try to bump some k8s
dependencies.

This is a problem, as we need to bump them together, manually.

This should fix it by removing them all from dependabot.


evrardjp:
Vulnerable image of alpine. Nothing we can do here.

ckotzbauer (approved):

dholbach (approved):

PR #276, opened 2020-12-14
Temporarily workaround alpine issue

evrardjp:
Until a new alpine image is created, we should ensure the latest
packages are used, and therefore we should upgrade default
installed packages.

Without this patch, we'll have outdated and vulnerable packages
until a new 3.12 image is released.

This is a problem, as we'll publish broken images.

This should temporarily workaround it, at the expense of larger
images (contains package cache)

### DIFF
--- a/cmd/kured/Dockerfile
+++ b/cmd/kured/Dockerfile
@@ -1,4 +1,4 @@
 FROM alpine:3.12
-RUN apk add --no-cache ca-certificates tzdata
+RUN apk update --no-cache && apk upgrade --no-cache && apk add --no-cache ca-certificates tzdata
 COPY ./kured /usr/bin/kured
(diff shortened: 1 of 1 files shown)


ckotzbauer (approved):
Is the image really larger now? There is no package cache created when using the `--no-cache` (as you do) or you delete the files afterwards `rm -rf /var/cache/apk/*`.
Or am I wrong here? 😁

evrardjp:
> Is the image really larger now? There is no package cache created when using the `--no-cache` (as you do) or you delete the files afterwards `rm -rf /var/cache/apk/*`.
> Or am I wrong here? grin

No, you're right. It's just a tiny bit larger, I suppose it's because it doesn't clean up the previous package.

dholbach (approved):

PR #279, opened 2020-12-17
Add force-reboot after force-timeout duration has been exceeded

cnmcavoy:
Updated version of https://github.com/weaveworks/kured/pull/109 which passes in the force-timeout parameter into the kubectl drain helper configuration, utilizing Go context cancellation instead.

I also updated the helm chart to support passing the new parameters.


dholbach:

PR #287, opened 2021-01-08
Corrected README Setting a schedule configuration example

jack-education:
Corrected README `Setting a schedule` configuration example

Added = for the setting of the scheduling commands


dholbach (approved):
Thanks a lot for the PR!

PR #288, opened 2021-01-11
Refresh kind cluster versions

evrardjp:
Without this patch, we are using outdated images in kind cluster
setup.

This should fix it, by removing 1.17 cluster (which is not tested
anymore), and updating 1.19 images. We are now with the
latest images published on dockerhub, as of today.


evrardjp:
This is not necessary, but nice to have.

dholbach (approved):
LGTM - thanks JP!

PR #260, opened 2020-12-01
Refactor reboot blocks

evrardjp:
Without this patch, we rely on global state in many functions for
which we check the reboot blockers.

This is a problem, as it's harder to test.

This patch fixes it by refactoring the reboot blockers. This also
includes a first series of unit tests for our main.

PR #297, opened 2021-01-22
Parametrize reboot commands

evrardjp:


evrardjp:
This is on top of #260 . We can either merge it with 2 prs or just one.

kingdonb:
I was going to suggest adding a make target that runs the tests, but this might be complicated since the binary won't compile on macos, the tests can't be run without some kind of linux environment. It would be great if the tests you've added here can be hooked up to run as part of the CI.

evrardjp:
@kingdonb IIRC, the `build image` test is running `make` + `make image`

kingdonb:
The Dockerfile isn't running any test from what I can tell, and there is no sign of test in the Makefile at all. There are a few tests that look like they are meant to be run by hand, on kind, one that sets up the reboot sentinels and another that checks the reboots proceeding orderly. I like the unit tests I saw that you added, this is a great technique to make source code better.

The linux binary is cross-compiled in the Makefile and copied into the docker image, it would be possible to run the test inside a build if there was a build tool and perhaps refactor this to a multi-stage build?

I found a Linux machine to run the tests on that you added:

```
$ cd cmd/kured/
$ go test
WARN[0000] Reboot blocked: prometheus query error: Post "/api/v1/query": unsupported protocol scheme ""
FATA[0000] Error invoking sentinel command: fork/exec ./babar: no such file or directory
PASS
ok  	github.com/weaveworks/kured/cmd/kured	0.068s
```

I am not sure if it's possible to redirect those logs or capture somehow, since they are good logs that indicate success of the test, (but outwardly giving the appearance of bad news during the test execution is less than helpful for anyone running the tests and reading along for success output.)

I think this is a great PR and I am likely to give it my approval if some of this can be addressed

evrardjp:
Rebased.

evrardjp:
New features in, need adapting. WIP.

evrardjp:
This should do the trick.

evrardjp:
Rebased for 1.20.

evrardjp:
@kingdonb go tests fixed in this PR. New go tests added in https://github.com/weaveworks/kured/pull/326.

ckotzbauer (approved):
I hadn't time to test, but the code looks good!

evrardjp:
Well, the CI does a preliminary test and doesn't seem to complain too much :)

evrardjp:
I won't merge my own PR, so someone else should do ;)

PR #300, opened 2021-02-07
add podLabels parameter

t3mi:
Adds configuration of additional pod labels, like cost center.


evrardjp (approved):
I don't see a problem with that. WDYT, @ckotzbauer ?

ckotzbauer (approved):
Same here, thanks for your contribution :+1:

dholbach (approved):

PR #325, opened 2021-03-19
Extend close-duration for stale issues and prs

### LINKED ISSUE #323: Auto expiry of issues/PR using github bot is happening too fast
This project is going at a pace for which expiring issues/PR after 7 days is too fast.

We are moving slower than that, so we should allow more grace time for those PRs and issues.

If someone can move it to 3 weeks instead of 1 that would be lovely.

ckotzbauer:
Closes #323 

### DIFF
--- a/.github/workflows/periodics-daily.yaml
+++ b/.github/workflows/periodics-daily.yaml
@@ -19,6 +19,7 @@ jobs:
         stale-issue-label: 'no-issue-activity'
         stale-pr-label: 'no-pr-activity'
         exempt-issue-labels: 'good-first-issue'
+        days-before-close: 21
 
   check-docs-links:
     name: Check docs for incorrect links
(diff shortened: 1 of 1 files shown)


.github/workflows/periodics-daily.yaml
```diff
@@ -19,6 +19,7 @@ jobs:
         stale-issue-label: 'no-issue-activity'
         stale-pr-label: 'no-pr-activity'
         exempt-issue-labels: 'good-first-issue'
+        days-before-close: 21
```
evrardjp, line 22: Can you also update L12 to match, thank you in advance!

evrardjp (changes requested):
Small nit, else it's good!

.github/workflows/periodics-daily.yaml
ckotzbauer: Thanks, good catch!

evrardjp (approved):
LGTM

PR #326, opened 2021-03-22
Add go tests for PRs

evrardjp:
Without this patch, go test bugs can appear without getting caught.

This should fix it.

### DIFF
--- a/.github/workflows/on-pr.yaml
+++ b/.github/workflows/on-pr.yaml
@@ -4,6 +4,20 @@ on:
   push:
 
 jobs:
+  pr-gotest:
+    name: Run go tests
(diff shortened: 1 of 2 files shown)


evrardjp:
This should fail, as we have borken go tests.

evrardjp:
Working as expected!

evrardjp:
(I could probably remove the annotation bit, as I don't think it matters, as long as we run the go tests, I am fine.

ckotzbauer:
I think it's a bonus which could improve the review experience. We can try it and remove it if we don't like this.

evrardjp:
Rebased. go tests should now work.

ckotzbauer (approved):

PR #315, opened 2021-03-03
Implement universal notification mechanism (NEW)

atighineanu:
Without this patch, kured can only notify via slack.

 This patch gives the possibility to send notifications
 across different technologies. Also, this patch makes
 slack-hook-url, slack-username and slack-channel
 deprecated (informed by a warning).
 Also, updated the documentation (Readme).

PR #333, opened 2021-03-30
update chart definition to include --notify-url

atighineanu:
Updating chart definition for --notify-url, following the PR: https://github.com/weaveworks/kured/pull/315

### DIFF
--- a/charts/kured/Chart.yaml
+++ b/charts/kured/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: "1.6.1"
+appVersion: "1.7.0"
 description: A Helm chart for kured
 name: kured
-version: 2.4.0
+version: 2.5.0
 home: https://github.com/weaveworks/kured
 maintainers:
   - name: ckotzbauer
--- a/charts/kured/README.md
+++ b/charts/kured/README.md
@@ -56,6 +56,7 @@ The following changes have been made compared to the stable chart:
 | `configuration.slackChannel` | cli-parameter `--slack-channel`                                        | `""`                      |
 | `configuration.slackHookUrl` | cli-parameter `--slack-hook-url`                                       | `""`                      |
 | `configuration.slackUsername` | cli-parameter `--slack-username`                                      | `""`                      |
+| `configuration.notifyUrl` | cli-parameter `--notify-url`                                              | `""`                      |
 | `configuration.messageTemplateDrain` | cli-parameter `--message-template-drain`                       | `""`                      |
 | `configuration.messageTemplateReboot` | cli-parameter `--message-template-reboot`                     | `""`                      |
 | `configuration.startTime` | cli-parameter `--start-time`                                              | `""`                      |
--- a/charts/kured/templates/daemonset.yaml
+++ b/charts/kured/templates/daemonset.yaml
@@ -88,6 +88,9 @@ spec:
           {{- if .Values.configuration.slackUsername }}
             - --slack-username={{ .Values.configuration.slackUsername }}
           {{- end }}
+          {{- if .Values.configuration.notifyUrl }}
+            - --notify-url={{ .Values.configuration.notifyUrl }}
+          {{- end }}
           {{- if .Values.configuration.messageTemplateDrain }}
             - --message-template-drain={{ .Values.configuration.messageTemplateDrain }}
           {{- end }}
--- a/charts/kured/values.minikube.yaml
+++ b/charts/kured/values.minikube.yaml
@@ -15,6 +15,7 @@ configuration:
   # slackChannel: ""          # slack channel for reboot notfications
   # slackHookUrl: ""          # slack hook URL for reboot notfications
   # slackUsername: ""         # slack username for reboot notfications (default "kured")
+  # notifyUrl: ""             # notification URL with the syntax as follows: https://containrrr.dev/shoutrrr/services/overview/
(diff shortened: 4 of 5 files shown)


charts/kured/README.md
```diff
@@ -56,6 +56,7 @@ The following changes have been made compared to the stable chart:
 | `configuration.slackChannel` | cli-parameter `--slack-channel`                                        | `""`                      |
 | `configuration.slackHookUrl` | cli-parameter `--slack-hook-url`                                       | `""`                      |
 | `configuration.slackUsername` | cli-parameter `--slack-username`                                      | `""`                      |
+| `configuration.notifyURL` | cli-parameter `--notify-url`                                              | `""`                      |
```
ckotzbauer, line 59: Small nit: Can you rename the parameter from "notifyURL" to "notifyUrl" to be consistent with e.g. "slackHookUrl"?

ckotzbauer (changes requested):
Thanks for the PR @atighineanu!
This LGTM if the naming is addressed properly 😉

ckotzbauer:
The merge of this PR has to wait until 1.7.0 is released.

charts/kured/README.md
atighineanu: Sure. I thought I'd keep the name consistent with the one in the main.go code.

charts/kured/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.6.1"
```
yershalom, line 2: Should the `appVersion` changed too?

ckotzbauer: No not now. This will be done when we make the new kured release.

evrardjp: As you mentioned: This has to wait for the 1.7.0 release to merge, so maybe we can bump it to "1.7.0" already, though it will certainly fail until 1.7.0 is public. Which seems fair to me. :)

evrardjp (commented):
Looks okay. I would already bump to 1.7.0, and wait for it to pass tests before merging.

But I have no strong opinion, I trust @ckotzbauer for this! :)

ckotzbauer:
I'm also fine with what @evrardjp suggests to increase the `appVersion` to 1.7.0 already, as this wouldn't be merged until it is released. So maybe that's the more elegant solution 😉

charts/kured/Chart.yaml
atighineanu: Done.

evrardjp:
I suppose the image also needs to be bumped to 1.7.0. It's currently located in Values.image.tag.

evrardjp:
@ckotzbauer can we, by default, make sure that the image tag is set to the app version? That would make it simpler.

ckotzbauer:
@evrardjp Yes, that's a good idea. I make a PR for that.

charts/kured/templates/daemonset.yaml
```diff
@@ -88,6 +88,9 @@ spec:
           {{- if .Values.configuration.slackUsername }}
             - --slack-username={{ .Values.configuration.slackUsername }}
           {{- end }}
+          {{- if .Values.configuration.notifyURL }}
+            - --notify-url={{ .Values.configuration.notifyURL }}
```
ckotzbauer, line 92: Please use `notifyUrl` here.

charts/kured/values.minikube.yaml
```diff
@@ -15,6 +15,7 @@ configuration:
   # slackChannel: ""          # slack channel for reboot notfications
   # slackHookUrl: ""          # slack hook URL for reboot notfications
   # slackUsername: ""         # slack username for reboot notfications (default "kured")
+  # notifyURL: ""              # notification URL with the syntax as follows: https://containrrr.dev/shoutrrr/services/overview/
```
ckotzbauer, line 18: Please use `notifyUrl` here.

evrardjp:
@ckotzbauer I wasn't sure if that was possible, but if it is, that's awesome! Maybe we can edit the Makefile to remove that sed in values.yaml then :)

charts/kured/values.yaml
```diff
@@ -34,6 +34,7 @@ configuration:
   slackChannel: ""           # slack channel for reboot notfications
   slackHookUrl: ""           # slack hook URL for reboot notfications
   slackUsername: ""          # slack username for reboot notfications (default "kured")
+  notifyURL: ""              # notification URL with the syntax as follows: https://containrrr.dev/shoutrrr/services/overview/
```
ckotzbauer, line 37: Please use `notifyUrl` here.

charts/kured/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: "1.6.1"
+appVersion: "1.7.0"
 description: A Helm chart for kured
 name: kured
-version: 2.4.0
+version: 2.4.1
```
evrardjp, line 5: This will be conflicting on rebase. Instead of bumping to 2.4.1, I would bump the version to 2.5.0, as we are moving to a new version of kured in practice.

evrardjp (changes requested):
See comment above.

ckotzbauer (approved):

evrardjp:
@atighineanu did you rebase before updating the value? It shows as still conflicting.

dholbach:

atighineanu:
> Sorry. Can you please retarget your branch to be merged into `main` please?

Sure.

PR #334, opened 2021-04-01
Fix invoke reboot for custom commands

evrardjp:
Without this patch, the rebootCommand passed to invokeReboot is
ignored, and the command used for reboot is always systemctl reboot.

This is a problem, as we are aiming for flexible commands for this
release.

This fixes it by restoring the previous behaviour before commit
[1] happened.

[1]: https://github.com/weaveworks/kured/commit/694957d56e7729ba6493fadca6cd9c1ac0a398fa

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/google/shlex"
 
+	shoutrrr "github.com/containrrr/shoutrrr"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/weaveworks/kured/pkg/alerts"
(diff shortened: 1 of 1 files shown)


evrardjp:
We should probably unit test those functions better.

atighineanu (approved):
LGTM

jackfrancis:
I think we can remove the `rebootCommand` global string var here: https://github.com/weaveworks/kured/pull/334/files#diff-277a207c771b9236363f103e013734989abd0a7f7f0a52ad86a6b1d32e6cfa10R58

evrardjp:
> 
> 
> I think we can remove the `rebootCommand` global string var here: https://github.com/weaveworks/kured/pull/334/files#diff-277a207c771b9236363f103e013734989abd0a7f7f0a52ad86a6b1d32e6cfa10R58

AFAIK it's still used to receive the arg from the CLI. 

Also, I would prefer doing a cleanup in a separate PR, if you don't mind. There are quite a few cleanups we could do :)

jackfrancis:
@evrardjp yep you're right, nevermind

lgtm

jackfrancis (approved):

ckotzbauer (approved):

PR #335, opened 2021-04-02
Use chart appVersion as default image-tag

ckotzbauer:
- Remove `sed` command for `values.yaml` in Makefile.
- Set tag in `values.yaml` to empty string


evrardjp (approved):
Makes things simpler, I like this!

Issue #332, opened 2021-03-30
Expose changes of #315 in Helm chart


dholbach:
Have a look at #315 and expose changes in Helm chart please.

atighineanu:
I am on it.

atighineanu:
Here is the PR:
https://github.com/weaveworks/kured/pull/333

evrardjp:
That will probably have to be _post_ 1.7.0. Do we need to update the manifest building for 1.7.0, though, to expose this feature?

ckotzbauer:
#338 is merged

PR #338, opened 2021-04-06
update chart definition to include --notify-url

atighineanu:
Updating chart definition for --notify-url, following the PR: #315
Issue(s): https://github.com/weaveworks/kured/issues/332


ckotzbauer (approved):

evrardjp (approved):
LGTM. Will only pass on releasing, but it's good already.

Let's hope github will be smart enough to do rebasing on its own if any changes come up before this one merges :)

evrardjp:
That's interesting, how is this passing? : )

evrardjp:
oh yeah it makes sense, the image used is the image for which the code was built, so HEAD of this branch. 
I wonder if, for helm chart changes, we should have a different behaviour, and only test existing images?

atighineanu:
> ~oh yeah it makes sense, the image used IS the image directly built on this PR (HEAD of this branch). Because the CI overrides the image used, this passes.~
> 
> ~I wonder if, for helm chart changes, we should have a different behaviour, and only test existing images? It's maybe too convoluted...~
> 
> Edit: I have noticed our chart testing action should have failed. Its logs show it's not working as expected, but the job is still green. Next to this, I have proposed a test coverage change to be more clear, as what I mentioned above is still true, but isn't obvious to me, even if I was the original author of that testing (I guess my memory is becoming bad with time >< )

Related PR: https://github.com/weaveworks/kured/pull/348

PR #339, opened 2021-04-06
Update github actions

evrardjp:
Without this patch, github actions are lagging behind.
This should improve our coverage.


dholbach (approved):
Nice work!

PR #340, opened 2021-04-06
Update dev docs

dholbach:
Some file names changed, so I went and fixed the dev docs.


evrardjp (changes requested):
Can we separate the go.mod and go.sum changes from this PR?

I don't see why they are bundled in this PR.

dholbach:
Sure, I can - I just noticed that VSCode was complaining that the `go.*` were out of sync...........

evrardjp:
> 
> 
> Sure, I can - I just noticed that VSCode was complaining that the `go.*` were out of sync...........

haha, same for me :)

evrardjp (approved):
LGTM

PR #341, opened 2021-04-06
Add force-reboot after force-timeout duration has been exceeded

cnmcavoy:
Resserrected copy of previous pull request https://github.com/weaveworks/kured/pull/279

---

Updated version of #109 which passes in the force-timeout parameter into the kubectl drain helper configuration, utilizing Go context cancellation instead.

I also updated the helm chart to support passing the new parameters.

### DIFF
--- a/charts/kured/Chart.yaml
+++ b/charts/kured/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.6.1"
 description: A Helm chart for kured
 name: kured
-version: 2.4.3
+version: 2.4.2
 home: https://github.com/weaveworks/kured
 maintainers:
   - name: ckotzbauer
--- a/charts/kured/Chart.yaml
+++ b/charts/kured/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.6.1"
 description: A Helm chart for kured
 name: kured
-version: 2.4.2
+version: 2.4.3
 home: https://github.com/weaveworks/kured
 maintainers:
   - name: ckotzbauer
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -38,27 +38,28 @@ var (
 	version = "unreleased"
 
 	// Command line flags
-	forceReboot               bool
-	drainTimeout              time.Duration
-	period                    time.Duration
-	drainGracePeriod          int
-	dsNamespace               string
-	dsName                    string
-	lockAnnotation            string
-	lockTTL                   time.Duration
-	prometheusURL             string
-	preferNoScheduleTaintName string
-	alertFilter               *regexp.Regexp
-	rebootSentinelFile        string
-	rebootSentinelCommand     string
-	notifyURL                 string
-	slackHookURL              string
-	slackUsername             string
-	slackChannel              string
-	messageTemplateDrain      string
-	messageTemplateReboot     string
-	podSelectors              []string
-	rebootCommand             string
+	forceReboot                     bool
+	drainTimeout                    time.Duration
+	period                          time.Duration
+	drainGracePeriod                int
+	skipWaitForDeleteTimeoutSeconds int
+	dsNamespace                     string
+	dsName                          string
+	lockAnnotation                  string
+	lockTTL                         time.Duration
+	prometheusURL                   string
+	preferNoScheduleTaintName       string
+	alertFilter                     *regexp.Regexp
+	rebootSentinelFile              string
+	rebootSentinelCommand           string
+	notifyURL                       string
+	slackHookURL                    string
+	slackUsername                   string
+	slackChannel                    string
+	messageTemplateDrain            string
+	messageTemplateReboot           string
+	podSelectors                    []string
+	rebootCommand                   string
 
 	rebootDays    []string
 	rebootStart   string
@@ -95,9 +96,11 @@ func main() {
 		Run:    root}
 
 	rootCmd.PersistentFlags().BoolVar(&forceReboot, "force-reboot", false,
-		"force a reboot even if the drain is still running (default false)")
+		"force a reboot even if the drain is still running (default: false)")
 	rootCmd.PersistentFlags().IntVar(&drainGracePeriod, "drain-grace-period", -1,
-		"grace period of time for pods to wait for the node drain in seconds (default -1)")
+		"time in seconds given to each pod to terminate gracefully, if negative, the default value specified in the pod will be used (default: -1)")
+	rootCmd.PersistentFlags().IntVar(&skipWaitForDeleteTimeoutSeconds, "skip-wait-for-delete-timeout", 0,
+		"when seconds is greater than zero, skip waiting for the pods whose deletion timestamp is older than N seconds while draining a node (default: 0)")
 	rootCmd.PersistentFlags().DurationVar(&drainTimeout, "drain-timeout", 0,
 		"timeout after which the drain is aborted (default: 0, infinite time)")
 	rootCmd.PersistentFlags().DurationVar(&period, "period", time.Minute*60,
@@ -341,15 +344,16 @@ func drain(client *kubernetes.Clientset, node *v1.Node) {
 	}
 
 	drainer := &kubectldrain.Helper{
-		Client:              client,
-		Ctx:                 context.Background(),
-		GracePeriodSeconds:  drainGracePeriod,
-		Force:               true,
-		DeleteEmptyDirData:  true,
-		IgnoreAllDaemonSets: true,
-		ErrOut:              os.Stderr,
-		Out:                 os.Stdout,
-		Timeout:             drainTimeout,
+		Client:                          client,
+		Ctx:                             context.Background(),
+		GracePeriodSeconds:              drainGracePeriod,
+		SkipWaitForDeleteTimeoutSeconds: skipWaitForDeleteTimeoutSeconds,
+		Force:                           true,
+		DeleteEmptyDirData:              true,
+		IgnoreAllDaemonSets:             true,
+		ErrOut:                          os.Stderr,
+		Out:                             os.Stdout,
+		Timeout:                         drainTimeout,
 	}
 
 	if err := kubectldrain.RunCordonOrUncordon(drainer, node, true); err != nil {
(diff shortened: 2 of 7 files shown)


cnmcavoy:
@dholbach I rebased and retarged the PR. It's unclear to me why the github action is failing, and I didn't find any documents on how to reproduce locally, so :\

evrardjp:
> 
> 
> @dholbach I rebased and retarged the PR. It's unclear to me why the github action is failing, and I didn't find any documents on how to reproduce locally, so :\

The command to run is `ct lint --config .github/ct.yaml` . Weirdly, this PR is not the first one to fail since our branch change.

evrardjp:
> 
> 
> > @dholbach I rebased and retarged the PR. It's unclear to me why the github action is failing, and I didn't find any documents on how to reproduce locally, so :\
> 
> The command to run is `ct lint --config .github/ct.yaml` . Weirdly, this PR is not the first one to fail since our branch change.

See also #344  for the fix to that test.

dholbach:
Thanks for rebasing and hanging in there - I hope we have things fixed up quickly again.

dholbach:
#344 is merged now - can you rebase again?

dholbach:
I'm not necessarily the best person to review this, but I pinged folks on Slack. Thanks a lot for rebasing!

charts/kured/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.6.1"
 description: A Helm chart for kured
 name: kured
-version: 2.4.2
+version: 2.4.3
```
jackfrancis, line 5: The convention is to make chart changes *after* this functional PR has merged, so take a stash of these changes (all changes under `charts/`) and have them ready to submit as a discrete PR after the changes to `cmd/kured/main.go` merge.

cmd/kured/main.go
```diff
@@ -333,19 +343,27 @@ func drain(client *kubernetes.Clientset, node *v1.Node) {
...
 		DeleteEmptyDirData:  true,
 		IgnoreAllDaemonSets: true,
 		ErrOut:              os.Stderr,
 		Out:                 os.Stdout,
 	}
+	if forceReboot {
+		drainer.Timeout = forceTimeout
+	}
+
 	if err := kubectldrain.RunCordonOrUncordon(drainer, node, true); err != nil {
```
jackfrancis, line 358: Wouldn't `--force-reboot` imply that we want to reboot even if cordon fails?

cmd/kured/main.go
```diff
@@ -38,7 +38,11 @@ var (
 	version = "unreleased"
 
 	// Command line flags
+	forceDrain                bool
+	forceReboot               bool
+	forceTimeout              time.Duration
```
jackfrancis, line 43: My 2¢ are that a more explicit "drainTimeout" option makes the most sense here. That way it can be used more generally, whether or not folks want to force a reboot if the timeout expires.

cmd/kured/main.go
```diff
@@ -38,7 +38,11 @@ var (
 	version = "unreleased"
 
 	// Command line flags
+	forceDrain                bool
```
jackfrancis, line 41: this "forceDrain" option looks to be unused

cmd/kured/main.go
cnmcavoy: The control flow here is a bit hard to follow here, but my understanding is that setting the `drainer.Timeout` field causes the `kubectldrain.RunNodeDrain` to err after the timeout has expired. It's a bit hard to spot, but the `log.Fatalf` does the actual work of panicing normally, and so setting `forceReboot` bypasses this panic, and so `drain` returns and `invokeReboot` is called.

cmd/kured/main.go
cnmcavoy: Good idea.

charts/kured/Chart.yaml
cnmcavoy: Done.

cmd/kured/main.go
jackfrancis: I'm talking about what to do in the err condition for invoking `kubectldrain.RunCordonOrUncordon`. I think `--force-reboot` should indicate that we want to reboot even if the node was not able to be marked as unschedulable, similar to how we are rebooting even if we aren't able to drain the node.

Basically, I think `--force-reboot` makes the most sense if *both* cordon and drain are best-effort, but not strictly required.

cnmcavoy: Oh I see. That makes sense to me. I will update the condition here to also not panic if force-reboot is set.

cmd/kured/main.go
```diff
@@ -91,6 +94,12 @@ func main() {
 		PreRun: flagCheck,
 		Run:    root}
 
+	rootCmd.PersistentFlags().BoolVar(&forceReboot, "force-reboot", false,
+		"force a reboot even if the drain is still running (default false)")
+	rootCmd.PersistentFlags().IntVar(&drainGracePeriod, "drain-grace-period", -1,
+		"grace period of time for pods to wait for the node drain in seconds (default -1)")
+	rootCmd.PersistentFlags().DurationVar(&drainTimeout, "drain-timeout", time.Minute*30,
```
jackfrancis, line 101: Are we following any existing guidance here on setting 30 mins as the default drain timeout?

cnmcavoy: Not exactly sure where that value came from. Looking at the code, the default should be zero (infinite).

cmd/kured/main.go
```diff
@@ -91,6 +94,12 @@ func main() {
 		PreRun: flagCheck,
 		Run:    root}
 
+	rootCmd.PersistentFlags().BoolVar(&forceReboot, "force-reboot", false,
+		"force a reboot even if the drain is still running (default false)")
+	rootCmd.PersistentFlags().IntVar(&drainGracePeriod, "drain-grace-period", -1,
+		"grace period of time for pods to wait for the node drain in seconds (default -1)")
+	rootCmd.PersistentFlags().DurationVar(&drainTimeout, "drain-timeout", 0,
```
jackfrancis, line 101: In theory I'm fine with a zero drain timeout, as that is operationally equivalent to the current implementation (no timeout specified).

I think in the future it might be sensible to express an opinion here, as a node "stuck" in a drain operation can slow down kured-induced node reboots across the cluster.

cmd/kured/main.go
```diff
@@ -333,19 +342,28 @@ func drain(client *kubernetes.Clientset, node *v1.Node) {
...
 		Out:                 os.Stdout,
+		Timeout:             drainTimeout,
 	}
+
 	if err := kubectldrain.RunCordonOrUncordon(drainer, node, true); err != nil {
-		log.Fatalf("Error cordonning %s: %v", nodename, err)
+		if !forceReboot {
+			log.Fatalf("Error cordonning %s: %v", nodename, err)
+		}
+		log.Errorf("Error cordonning %s: %v, continuing with reboot anyway", nodename, err)
```
evrardjp, line 359: Is there a side effect possible:
- when forceReboot
- when the cordon has failed
- and then the drain is triggered ?

cmd/kured/main.go
```diff
@@ -91,6 +94,12 @@ func main() {
 		PreRun: flagCheck,
 		Run:    root}
 
+	rootCmd.PersistentFlags().BoolVar(&forceReboot, "force-reboot", false,
+		"force a reboot even if the drain is still running (default false)")
+	rootCmd.PersistentFlags().IntVar(&drainGracePeriod, "drain-grace-period", -1,
+		"grace period of time for pods to wait for the node drain in seconds (default -1)")
```
evrardjp, line 100: it would best to clarify what -1 means: Wait for pod's terminationGracePeriodSeconds (like you rightfully did in drain-timeout)

cmd/kured/main.go
evrardjp: I agree with @jackfrancis . I would prefer set that in a different PR though, and in a different release (we have many things lined up for this one).

evrardjp: Nice that you mention "infinite time", it's definitely good for our users!

cmd/kured/main.go
```diff
@@ -91,6 +94,12 @@ func main() {
 		PreRun: flagCheck,
 		Run:    root}
 
+	rootCmd.PersistentFlags().BoolVar(&forceReboot, "force-reboot", false,
+		"force a reboot even if the drain is still running (default false)")
+	rootCmd.PersistentFlags().IntVar(&drainGracePeriod, "drain-grace-period", -1,
```
evrardjp, line 99: Should we also expose SkipWaitForDeleteTimeoutSeconds ?

evrardjp (commented):
I like where this is heading

cmd/kured/main.go
cnmcavoy: Sure. I added it to this PR.

cmd/kured/main.go
cnmcavoy: Updated to clarify what happens when the value is negative.

cmd/kured/main.go
cnmcavoy: it's unlikely, but just to be safe, I added an explicit return after the `Errorf` log statement.

evrardjp: Sorry this is not what I meant.

I meant that `drain` itself could be refactored to bubble up errors if necessary. Instead of having the current logic:
`if not force, fatal, else log error` in the current code, we could make sure drain returns an error, and the `rebotoAsRequired` could then behave differently, based on the error.

FYI it's a nit. It's food for thought. You don't need to change your current code, as it does what you tell it does.
What I mentioned can be done in a later patch. I just think it's probably easier to read that way. WDYT?

cnmcavoy: > Sorry this is not what I meant. I meant that I don't really know the details behind this kubernetes behaviour.

Oh I see. I assumed you meant draining when the cordon failed could lead to unexpected states.

I agree the error handling is all over the place here. The `forceReboot` condition makes the error handling even harder to follow. I'm not against rewriting `drain` to return an error if you think its in the scope of this PR.

evrardjp: Let's do that in another PR :) 

I would be happy to see a follow up PR from you!

evrardjp (approved):
I find that good enough. We can iterate later on follow up PRs (expose the feature to helm chart, refactorings)

evrardjp:
I am leaving the time for others to review, if nobody has reviewed in the next days, I will merge this.

evrardjp:
Nobody opposes, let's merge!

PR #343, opened 2021-04-07
DNM: Test helm chart

evrardjp:
DNM: Testing PR - test helm chart pipeline

### DIFF
--- a/.github/ct.yaml
+++ b/.github/ct.yaml
@@ -1,5 +1,6 @@
 # See https://github.com/helm/chart-testing#configuration
 remote: origin
+target-branch: main
 chart-dirs:
   - charts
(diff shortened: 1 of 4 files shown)


ckotzbauer:
@evrardjp There is a "target-branch" Setting of the "chart-testing" tool which is set to "master" by default. We have to change this in the ct.yaml

ckotzbauer:
https://github.com/helm/chart-testing/blob/master/ct/cmd/root.go#L66

evrardjp:
Oh right!

The problem I have seen on my local environment is the failure to get a folder, so I followed that. But indeed that can be because he couldn't find the branch at all. me is a dumbdumb.

PR #344, opened 2021-04-07
Add chart-testing target-branch

evrardjp:
Without this patch, chart-testing is using the branch named
"master" by default.

This is a problem, as we just renamed our development branch
"main" instead of "master".

This should fix it by pointing to the right branch.


dholbach (approved):

ckotzbauer (approved):
I will merge this, when all tests are green.

PR #345, opened 2021-04-07
Update deps

evrardjp:
This is the result of a go mod tidy.
It should clarify our dependencies.


dholbach (approved):

ckotzbauer (approved):

evrardjp:
We probably have other deps to update, but that wasn't the first scope of this commit. Let's keep those deps bumps separate :)

PR #283, opened 2020-12-29
add a drain timeout

flbla:
same as https://github.com/weaveworks/kured/pull/79 but with latest changes

Closes: #78

PR #351, opened 2021-04-12
add a drain timeout

flbla:
issue : #78 
rebase on main branch instead of master (previous PR : #283)

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -39,6 +39,7 @@ var (
 
 	// Command line flags
 	period                    time.Duration
+	drainTimeout              time.Duration
 	dsNamespace               string
 	dsName                    string
 	lockAnnotation            string
@@ -56,6 +57,7 @@ var (
 	messageTemplateReboot     string
 	podSelectors              []string
 	rebootCommand             string
+	isDrained                 bool
 
 	rebootDays    []string
 	rebootStart   string
@@ -93,6 +95,8 @@ func main() {
 
 	rootCmd.PersistentFlags().DurationVar(&period, "period", time.Minute*60,
 		"sentinel check period")
+	rootCmd.PersistentFlags().DurationVar(&drainTimeout, "drain-timeout", 0,
+		"drain timeout")
 	rootCmd.PersistentFlags().StringVar(&dsNamespace, "ds-namespace", "kube-system",
 		"namespace containing daemonset on which to place lock")
 	rootCmd.PersistentFlags().StringVar(&dsName, "ds-name", "kured",
@@ -315,7 +319,7 @@ func release(lock *daemonsetlock.DaemonSetLock) {
 	}
 }
 
-func drain(client *kubernetes.Clientset, node *v1.Node) {
+func drain(client *kubernetes.Clientset, node *v1.Node) bool {
 	nodename := node.GetName()
 
 	log.Infof("Draining node %s", nodename)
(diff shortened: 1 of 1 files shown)


flbla:
@evrardjp @dholbach

evrardjp:
Awesome! Thankls @flbla

cmd/kured/main.go
```diff
@@ -339,14 +343,17 @@ func drain(client *kubernetes.Clientset, node *v1.Node) {
...
 		Out:                 os.Stdout,
+		Timeout:             drainTimeout,
 	}
 	if err := kubectldrain.RunCordonOrUncordon(drainer, node, true); err != nil {
 		log.Fatalf("Error cordonning %s: %v", nodename, err)
 	}
 
 	if err := kubectldrain.RunNodeDrain(drainer, nodename); err != nil {
-		log.Fatalf("Error draining %s: %v", nodename, err)
+		log.Error("Error draining %s: %v", nodename, err)
```
evrardjp, line 353: This should probably be Errorf

cmd/kured/main.go
```diff
@@ -531,11 +538,21 @@ func rebootAsRequired(nodeID string, rebootCommand []string, sentinelCommand []s
...
-		drain(client, node)
-		invokeReboot(nodeID, rebootCommand)
-		for {
-			log.Infof("Waiting for reboot")
-			time.Sleep(time.Minute)
+		isDrained = drain(client, node)
+		if isDrained {
+			invokeReboot(nodeID, rebootCommand)
+			for {
+				log.Infof("Waiting for reboot")
```
evrardjp, line 545: this should probably be `log.Info`

cmd/kured/main.go
```diff
@@ -531,11 +538,21 @@ func rebootAsRequired(nodeID string, rebootCommand []string, sentinelCommand []s
...
+				time.Sleep(time.Minute)
+			}
+		} else {
+			log.Infof("Uncordon %s", node.GetName())
+			uncordon(client, node)
+			deleteFlag := newCommand("/usr/bin/nsenter", "-m/proc/1/ns/mnt", "/bin/rm", rebootSentinelFile)
```
evrardjp, line 551: I think it would be best to use the available functions for wrapping things properly (so that it works on all OSes).

Next, what about the cases where the sentinel is a command instead of a sentinel file?

cmd/kured/main.go
```diff
@@ -339,14 +343,17 @@ func drain(client *kubernetes.Clientset, node *v1.Node) {
...
+		Timeout:             drainTimeout,
 	}
 	if err := kubectldrain.RunCordonOrUncordon(drainer, node, true); err != nil {
 		log.Fatalf("Error cordonning %s: %v", nodename, err)
 	}
 
 	if err := kubectldrain.RunNodeDrain(drainer, nodename); err != nil {
-		log.Fatalf("Error draining %s: %v", nodename, err)
+		log.Error("Error draining %s: %v", nodename, err)
+		return false
```
evrardjp, line 354: There is another PR that grew a bit, and is touching this, maybe it's worth sharing efforts? 

Can you have a look at https://github.com/weaveworks/kured/pull/341/files ?

evrardjp (changes requested):
A few changes might be required, and this might conflict with an existing PR. It might be worth merging the two approaches.

cmd/kured/main.go
flbla: I maybe missed something, but I think it doens't uncordon the nodes after the timeout in the #341 ?

evrardjp: Correct.

Here, I think error handling can be indeed improved. We can change the function signature to return the success AND the error.

If an error happens, log it. If forceReboot isn't true, we should probably uncordon and continue the main loop. Else, the control flow continues. That makes it far clearer to read, IMO.

fouadsemaan:
The docs claim this change was put in as part of release 1.7.0. Is there an ETA?

evrardjp:
@fouadsemaan  TBH, I am not sure this _PR_ is necessary anymore. It gives extra details compared to a merged feature.
I would love to hear @flbla and your opinion on this.

Because the other feature was merged already, you might be interested by it @fouadsemaan .

flbla:
Hi,
I'm not sure if a drain timeout happened if it actually uncordon nodes
If not, this PR is still necessary

Issue #277, opened 2020-12-14
Feature request: Add a delay before rebooting another node


richerlariviere:
We have to support a service in kubernetes which doesn't tolerate node reboots very well. When we need to reboot a node, we must wait for our service to be stabilized before rebooting another node. I hope we could fix the service instead of trying to patch it around but it's not possible.

Something that could help us is a way to tell Kured to wait between node reboots. I think of a flag we could pass e.g.: `--delay-before-next-reboot 1h0m0s`.

I could eventually contribute to this feature but I still need to figure how to do it. Do you have any guidance or opinion about this issue?

Thanks a lot

rbq:
Same here. I'd like to space reboots a couple hours apart, just to give my distributed storage time to sync data that was written in the meantime.

jackfrancis:
@richerlariviere in the meantime you could set the `--period` runtime configuration value to "desired wait time" x 2. (You want to double it because your per-node cycles are not necessarily closely synchronized, so another could be ready to check for a reboot quickly after another node was rebooted.)

Have you tried that to see if it accomplishes the desired delay?

spingel:
We have tried the suggested work-around but observed that multiple nodes may still reboot within a short period. If agent pods are not started at the exact same time the update check runs at different times on different nodes. If one of the nodes completes a reboot and another node's update check happens to run shortly after this can cause pods across multiple nodes to be pending for some time.

+1 for adding a parameter to allow for delaying the release of the lock after a rebooted node was uncordoned.

evrardjp:
I need to mention that @spingel committed here ^ . If you are interested by this feature, please review #352 !

spingel:
Looks like this feature was released with 1.7. Thanks! Only outstanding change is the update for the helm chart: https://github.com/weaveworks/kured/pull/353 (I rebased the PR and updated the version).

PR #353, opened 2021-04-13
Add lockReleaseDelay parameter to helm chart

spingel:
Adds support for the --lock-release-delay parameter that was added in #352 to the helm chart. Setting `configuration.lockReleaseDelay=1h` for instance throttles node reboots to at most 1 per hour.

Issue: #277

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,7 @@ Flags:
   -h, --help                                help for kured
       --lock-annotation string              annotation in which to record locking node (default "weave.works/kured-node-lock")
       --lock-ttl duration                   expire lock annotation after this duration (default: 0, disabled)
+      --lock-release-delay duration         hold lock after reboot by this duration (default: 0, disabled)
       --message-template-drain string       message template used to notify about a node being drained (default "Draining node %s")
       --message-template-reboot string      message template used to notify about a node being rebooted (default "Rebooting node %s")
       --notify-url                          url for reboot notifications (cannot use with --slack-hook-url flags)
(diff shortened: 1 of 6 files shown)


evrardjp:
Note for reviewers: A bug in our CI can show everything green (Wolkswagen style!) in the tests, even the log tell you a failure. I am investigating this when I have the time.

ckotzbauer (approved):
LGTM. The `appVersion` has to change before merge, when we know in which kured-Version we will release this feature.
The failures @evrardjp mentioned are not releated to this PR, so I'll approve this.

evrardjp:
> 
> 
> Note for reviewers: A bug in our CI can show everything green for the helm chart install/lint (Wolkswagen style!) in the tests, even when the log indicates a failure. I am investigating this, but any help is welcomed, as I don't have much time for this.

This PR will start to fail when #355 will be merged.

EduardFazliev:
That's nice feature, we really want it, thanks for doing that. Can we use 2.5.0 version when it is going to be released with 1.16 k8s clusters?

evrardjp:
@EduardFazliev The support matrix is on our main README. For your case, kubernetes 1.16 was dropped with kured 1.5.1.
The feature we are talking about needs a higher version of kured, and therefore won't be available to you, in principle.
You can try running new versions of kured on old kubernetes clusters, but it's not tested and most likely a bad idea. YMMV.

ckotzbauer:
Can you add the new parameter to the `configuration` object in `values.yaml` as well? It seems I missed this in my first review. Then this is ready for merge. @spingel

ckotzbauer (approved):

PR #355, opened 2021-04-13
Reduce false positives

evrardjp:
Without this change, the "Test helm chart (install) action" will
rightfully succeed when our helm chart gets installed and has
no syntax issues. However, it doesn't test if kured is properly
installed. For example, the helm chart can try to install a
yet unpublished image, and our test will succeed, as the syntax
is still valid.

This is a problem, as everything looks green, but it's not
effectively working. Our other jobs are focusing on code changes,
so it effectively overrides the image tag, which is not what
we want in this "Test helm chart" action.

This fixes it by adding an extra command in the test.

### DIFF
--- a/.github/workflows/on-pr-charts.yaml
+++ b/.github/workflows/on-pr-charts.yaml
@@ -39,3 +39,35 @@ jobs:
 
       - name: Run chart tests
         run: ct ${{ matrix.test-action }} --config .github/ct.yaml
+
+  # This doesn't re-use the ct actions, due to many limitations (auto tear down, no real testing)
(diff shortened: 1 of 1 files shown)


evrardjp:
See also failure in https://github.com/weaveworks/kured/pull/356

evrardjp:
DNM yet. Caught an issue.

evrardjp:
I am not confident to merge this yet.

evrardjp:
I guess I was too careful. It's rightfully working! :)

dholbach (approved):
nice work!

PR #356, opened 2021-04-13
DNM test break

evrardjp:
- Reduce false positives in chart testing
- DNM Ensure this is broken


evrardjp:
This should break, and is just for demo purposes.

evrardjp:
Rightfully failing. This PR has fullfilled its purpose and can now be closed.

PR #357, opened 2021-04-14
Update kured-ds.yaml

evrardjp:
Without this patch, it's not clear that we added command line
arguments recently. This should expose our latest changes in the
future released manifest.

### DIFF
--- a/kured-ds.yaml
+++ b/kured-ds.yaml
@@ -44,23 +44,31 @@ spec:
                   fieldPath: spec.nodeName
           command:
             - /usr/bin/kured
-#            - --alert-filter-regexp=^RebootRequired$
-#            - --blocking-pod-selector=runtime=long,cost=expensive
-#            - --blocking-pod-selector=name=temperamental
-#            - --blocking-pod-selector=...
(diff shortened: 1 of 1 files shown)


evrardjp:
I re-ordered those to be equal to the order in the code.

I am thinking of reordering them by function or alphabetically, but I think that can be done later. As long as we don't miss anything.

dholbach:
Hum... why code-order? Not sure if that's what users expect?

evrardjp:
I used the code order so I didn't forget one line :) Those things are easy to miss.

I am happy to change the order to something that makes more sense to our users.
Do you have anything to suggest? It needs to be clear so that new contributors KNOW how to add their flag, if necessary.

evrardjp:
One way is around integrations: split the arguments around notifications, reboot blockers, reboot triggers. But I am not super convinced yet.

evrardjp:
Still, in the meantime, it's worth being more complete than what we currently have.

I mentioned this gap into the PR to clarify the contributions and test coverage.

ckotzbauer (approved):

PR #360, opened 2021-04-19
Add force-reboot and drain timeouts to chart config and ds

cnmcavoy:
Followup from https://github.com/weaveworks/kured/pull/341, exposes all of the new cli flags to the helm chart.

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -100,7 +100,7 @@ func main() {
 		Run:    root}
 
 	rootCmd.PersistentFlags().BoolVar(&forceReboot, "force-reboot", false,
-		"force a reboot even if the drain is still running (default: false)")
+		"force a reboot even if the drain fails or times out (default: false)")
 	rootCmd.PersistentFlags().IntVar(&drainGracePeriod, "drain-grace-period", -1,
 		"time in seconds given to each pod to terminate gracefully, if negative, the default value specified in the pod will be used (default: -1)")
 	rootCmd.PersistentFlags().IntVar(&skipWaitForDeleteTimeoutSeconds, "skip-wait-for-delete-timeout", 0,
--- a/charts/kured/Chart.yaml
+++ b/charts/kured/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.7.0"
 description: A Helm chart for kured
 name: kured
-version: 2.9.0
+version: 2.9.1
 home: https://github.com/weaveworks/kured
 maintainers:
   - name: ckotzbauer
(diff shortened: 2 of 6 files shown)


anthr76:
Would it be possible to add reboot command as well or is another PR warranted?

evrardjp:
@anthr76 the reboot command is already flexible, in our main branch. We are just waiting for the next kubernetes kind nodes to release our recent work!

cmd/kured/main.go
```diff
@@ -96,7 +96,7 @@ func main() {
 		Run:    root}
 
 	rootCmd.PersistentFlags().BoolVar(&forceReboot, "force-reboot", false,
-		"force a reboot even if the drain is still running (default: false)")
+		"force a reboot even if the drain fails or times out (default: false)")
```
evrardjp, line 99: if you would like this clarification to be included in the next release (1.7.0), it's probably better to add it into another PR.

cnmcavoy: I'm not overly concerned, I just noticed it was slightly incorrect when doing the helm chart changes and wanted to correct it.

bstrdsmkr:
> @anthr76 the reboot command is already flexible, in our main branch. We are just waiting for the next kubernetes kind nodes to release our recent work!

I'm not sure I understand your response here -- I believe @anthr76 is asking to have the `--reboot-command` included as an explicit option in the chart, even though it (as are these) is already settable via `extraArgs`. I didn't see an explicit option in the chart in the main branch. Did I just miss it?

anthr76:
@bstrdsmkr Is indeed correct.. Didn't see that `extraArgs` bit before so not totally necessary. Just a nice to have at this point.

evrardjp:
You're right, it seems like this slipped through the cracks. PRs welcome!

anthr76:
I'll be happy to. Any idea on when this get's merged?

evrardjp:
> 
> 
> as a sanity check, is there a generic way to convert between the flag name and a yaml safe name? That way you could generate them from the yaml values instead of hard coding each one

That would be amazing, but I am not sure we have that. @ckotzbauer what do you think of the idea? We could maybe write a utility function that does that in the helm chart, and leverage it? Or is there any other way?

evrardjp:
@cnmcavoy this will need a rebase, can you tackle it? That would be lovely.

ckotzbauer:
@evrardjp @bstrdsmkr I think the "kebabcase" function could help us: https://helm.sh/docs/chart_template_guide/function_list/#kebabcase

EduardFazliev:
Is there any chance that this will be merged some day? Is there a way to help to speed up the process?

ckotzbauer:
@cnmcavoy Can you rebase this again please?
I think this is should be ready to merge in general, if there are no other objections. I would be ok to include the clarification in `main.go`
/cc @evrardjp

cnmcavoy:
rebased.

ckotzbauer (approved):

PR #365, opened 2021-04-27
Use stable kind-action

evrardjp:
We are relying on master, which might break anytime (or in this
case, moved to another branch).

Instead we should rely on a stable version, and unfreeze if
necessary. Dependabot helps us maintain those releases anyway.


ckotzbauer (approved):

jackfrancis (commented):
definitely lgtm

PR #368, opened 2021-04-29
removed notifications/slack package 

atighineanu:
In this PR the slack-hook-url is translated
 into shoutrrr syntax. Therefore, slack pack
 age as well as checks for slack-hook-url in
 drain and reboot functions are removed.
 
 I suppose we will keep the --slack-hook-url for an undefined amount of releases. Therefore, I kept it while I got rid of the notifications/slack package.


evrardjp:
Look at all this red (in commit diff), and all this green (in CI). I vote for that! 
Can you rebase it, please?

cmd/kured/main.go
```diff
@@ -166,12 +165,20 @@ func main() {
...
+
+	if slackHookURL != "" {
 		log.Warnf("Deprecated flag(s). Please use --notify-url flag instead.")
+		tmpsplit := strings.Split(slackHookURL, "/")
+		if len(tmpsplit) < 4 {
+			log.Warnf("slack-hook-url is not properly formatted...\nno notification will be sent.")
+		} else {
+			tokenA := tmpsplit[len(tmpsplit)-3]
+			tokenB := tmpsplit[len(tmpsplit)-2]
+			tokenC := tmpsplit[len(tmpsplit)-1]
```
dholbach, line 178: Would it make sense to add a couple of obvious examples in a unit test to check that this works?

evrardjp: yes, I would extract a bit of that logic into a unit testable function, and add the tests. Agree with @dholbach . Or just unit test the flagCheck into one go, which sounds better... (However, we only have one series of cli flags checked in that form for now! We should probably check for the existence of the command for reboot or for sentinel, but this is unrelated to this PR...)

Next to that, I would add a comment before the L166 and L170 to tell that this code should be removed in the next release: We probably want to clean up this "automatic migration" in the future.

evrardjp (changes requested):
Needs a few changes and a rebase :)

atighineanu:
1) Unit test for flagCheck() function is added.
2) The code is improved (used net/url and strings.Trim instead of stupidly splitting pieces of the URL).

evrardjp:
That diff looks good. I will have a view in more details soon.

cmd/kured/main.go
```diff
@@ -173,12 +173,20 @@ func main() {
...
 		log.Warnf("Cannot use both --notify-url and --slack-hook-url flags. Kured will use --notify-url flag only...")
-		slackHookURL = ""
-	}
-	if slackChannel != "" || slackHookURL != "" {
-		log.Warnf("slack-* flag(s) are being deprecated. Please use --notify-url flag instead.")
 	}
 
+	if slackHookURL != "" {
+		log.Warnf("Deprecated flag(s). Please use --notify-url flag instead.")
+		trataURL, err := url.Parse(slackHookURL)
```
evrardjp, line 180: The more I think about this, the more I am inclined to say: "Don't ignore incorrect input. Instead crash to show the user has done something wrong". I know it's a change in our way of doing, but I would like us to think about it.

Which means, in this case:

- An error in parsing the slackHookUrl should return an error, and in caller function, we should handle that error by crashing.
- Side note: I think that the url could match a regex, and fail if it's not matching: (https://hooks.slack.com/services/(.*)/(.*)/(.*)), the result don't need to be tested for length, as the expectations are there. Again, returning an error would be more explicit.
- Alternative to previous idea, assign a slice after split(trim (path)). This would make L184 and L187 clearer.

evrardjp (commented):
Up for discussion :)

cmd/kured/main.go
atighineanu: I am not so sure regarding the crash if slack hook url is incorrect formatted. @dholbach, @ckotzbauer, @jackfrancis what say you?

atighineanu: I thought it htrough, I think the regexp.Mustcompile() thing would be obsolete because url.Parse() from net/url is already doing this. It will throw an error and the user will see it in the log. Question still remains: log.Printf or log.Fatalf ? 
Also, if there is "/" at the end of the URL, shoutrrr will throw an error which will be exposed in the log as well.

PR #382, opened 2021-05-28
fix slack deprecation notice

### LINKED ISSUE #381: Deprecation warning is always printed, even when slack-* cli args are not used
Hey folks 👋🏼 

Mostly a minor annoynace, but since the `slackUsername` flag has a non-empty default
https://github.com/weaveworks/kured/blob/main/cmd/kured/main.go#L134

it always triggers the deprecation notice:
https://github.com/weaveworks/kured/blob/5193f2de1666f50a0c8fd5906c534ef77e4a4ad4/cmd/kured/main.go#L171-L173

It would also be nice to indicate which flags are being deprecated in the warning message.

dkulchinsky:
fixes #381

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -47,7 +47,7 @@ var (
 	dsName                          string
 	lockAnnotation                  string
 	lockTTL                         time.Duration
-	lockReleaseDelay                time.Duration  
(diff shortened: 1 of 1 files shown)


dholbach (approved):
Thanks for the PR @dkulchinsky!

Issue #385, opened 2021-06-05
Filter Prometheus Alerts by regexp is limited, add feature to query by labels 


SimeonPoot:
Currently we'd like to query Prometheus for certain Active Alerts to block nodes from being rebooted, this functionality is present. It's great to have this functionality in place, However filtering by regex is limited in a way we can't use it. 

The situation is that we have different priorities of workloads getting the same alerts, that is; 
If teams are experimenting with their setup and it fires an alert, lets say '**PodCrashing**', we don't want to Kured to be blocked by this Alert.. But we do want to block Kured when this alert is fired on the workloads of the kubernetes-infra team (as this hits all teams). 
Rather than use different names for alerts per priority, it would be great if we could query Prometheus for active Alerts by 'labels'. This way we could utilize labels like: **"severity":"critical","team":"kubernetes-infra"** or something in line of this. 

Because we'd like to only block Kured on specific situations, I think it a hit on the query of labels should block Kured to reboot, thus doing the opposite of the regex. 

I'm not sure if this would be a good feature to introduce, but I've been working on implementing this in code ([with a bunch of  unit-tests](https://github.com/weaveworks/kured/issues/243)) and at the moment starting to test it on KiND. I'll link this to a PR #386 .

evrardjp:
Should we close this? If there is extra work to do, please mention it :)

SimeonPoot:
hi @evrardjp. thanks for coming back to this. 
The issue is still relevant. The only change that was made with the PR #386 , was restructuring of PrometheusClient as well as adding unit-tests. I removed all of its 'query by label' code as this needed some more investigation / better way of implementing. 
- I saw that there's an prometheus endpoint /alerts, where we could make use of, rather than /query. 
- The way of querying by labels: Investigate how Prometheus does this; would be great if the way the alerts are queried, is the same as Prometheus. A uniform way will reduce complexity. 

What do you think of this?

evrardjp:
As you can guess with the delay in my answers: I am not really available to work on this.
Hence, if you do the work, you are more likely to get things merged ;)

I am not really familiar with prometheus internals, but I think it's worth investigating, and documenting _WHY_ we want to change or NOT change. Example: Is the /alerts endpoing better? Why? Should we have a sidecar model instead, or should we keep this in code?

Agree on making things uniform, but not sure what you are proposing here.

SimeonPoot:
I can image, I'd like to work on it, but this comes between work/kids. 

[...]

Issue #243, opened 2020-11-27
Add unit tests for slack/prometheus behaviour


evrardjp:
As our automated (in github actions) release testing only does pure functional testing, we might miss some of the extra features kured is providing, like messaging notifications or prometheus integration.

To avoid issues creeping in into the release, we should probably test those integrations into each PR.
By testing those into PRs, we can also safely bump dependabot updates. (This could technically allow an auto-merge of dependabot dependencies in the future, as long as they are not client-go/kubectl related).

PR #386, opened 2021-06-05
Restructuring Prometheus client, added unit-tests to regex-queries active alerts

SimeonPoot:
This PR was made to satisfy a simple use-case to block Kured from rebooting a node, based on Prometheus labels. More detail described in this issue: #385. 

**However**, querying labels based on alerts needs to have some more thought, this because it could lead to possible breaking changes. That's why this has been removed from the PR. The Unit-tests and restructuring of the Prometheus Client is still there, because it's still relevant and it will improve extensibility and quality to the project. 

There will be a follow-up PR to introduce the change stated above. 

**This PR contains:** 
- Existing filter functionality is kept (regex), 
- Restructured Prometheus Client;
  - So we can do unit-testing,
  - Also get more flexibility to inject specific Prometheus configuration (if needed in the future),
  - Renamed PrometheusActiveAlerts function to ActiveAlerts method (of PromClient struct),
- Added Prometheus unit-tests with a mock-server, #243 

> Tested with unit-tests, and on a **KiND** cluster with connection to a Prometheus server with active alerts.


pkg/alerts/prometheus.go
```diff
@@ -7,42 +7,77 @@ import (
...
 				if alertName, isAlert := sample.Metric[model.AlertNameLabel]; isAlert && sample.Value != 0 {
 					if filter == nil || !filter.MatchString(string(alertName)) {
+						log.Info("Added to the BlockingList, by regexFilter: ", alertName)
 						activeAlertSet[string(alertName)] = true
 					}
+					for k, v := range inclLabel {
+						if sample.Metric[model.LabelName(k)] == model.LabelValue(v) {
+							// because filtering by labels doesn't check for alertname, we want to make sure
+							// there's no deadlock, that is, Kured is alerting itself on the block-list.
+							if alertName == "RebootRequired" {
```
evrardjp, line 66: what is this string? Should this be a const? I thought (proving my terrible understanding) that our prometheus gauge was reboot_required. Is the alert automatically named RebootRequired? Should we expose that to the user, or at least document it in the README?

cmd/kured/main.go
```diff
@@ -228,8 +228,8 @@ type RebootBlocker interface {
 // PrometheusBlockingChecker contains info for connecting
 // to prometheus, and can give info about whether a reboot should be blocked
 type PrometheusBlockingChecker struct {
-	// URL to contact prometheus API for checking alerts
-	promURL string
+	// labels to query get  alerts
+	includeLabel map[string]string
```
evrardjp, line 232: Alternatively, we can change this by splitting this into two blockers, as it's effectively what this is.
It duplicates a bit more code, but the separation of those are very clear, and the logic is easier to follow.

We would have to rename the current type as PrometheusAlertsBlockingChecker, and a new one to PrometheusLabelsBlockingChecker. While it would be easier to show/follow what is blocking, it's also quite a big overhead in terms of code, so I am really on the fence about my own proposition.

I hope I will get a community opinion on this comment ;)

pkg/alerts/prometheus_test.go
```diff
@@ -0,0 +1,216 @@
+package alerts
```
evrardjp, line 1: This is very nice.

evrardjp (commented):
I like where this is heading. Nice.
A few comments for my understanding.

DEVELOPMENT.md
```diff
@@ -171,6 +171,9 @@ A test-run with `kind` could look like this:
 # create kind cluster
 kind create cluster --config .github/kind-cluster-<k8s-version>.yaml
 
+# load the image to the kind cluster
```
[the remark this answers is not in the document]
SimeonPoot: I added this because I did this when I was testing. But I see your point, lets keep it clean and let the docs focus on a global procedure. I will remove this comment from the docs.

cmd/kured/main.go
```diff
@@ -228,8 +228,8 @@ type RebootBlocker interface {
 // PrometheusBlockingChecker contains info for connecting
 // to prometheus, and can give info about whether a reboot should be blocked
 type PrometheusBlockingChecker struct {
-	// URL to contact prometheus API for checking alerts
-	promURL string
+	// labels to query get  alerts
```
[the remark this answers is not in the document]
SimeonPoot: The Url will be configured when instantiating the prometheusClient, so no need to have it here, I guess?
Otherwise, we could add the PrometheusClient to this struct and use it? just like the "KubernetesBlockingChecker  struct. 
I'm not sure what to do here, if I'm being honest.

cmd/kured/main_test.go
```diff
@@ -23,7 +23,7 @@ func Test_rebootBlocked(t *testing.T) {
 	noCheckers := []RebootBlocker{}
 	nonblockingChecker := BlockingChecker{blocking: false}
 	blockingChecker := BlockingChecker{blocking: true}
-	brokenPrometheusClient := PrometheusBlockingChecker{promURL: "", filter: nil}
+	brokenPrometheusClient := PrometheusBlockingChecker{filter: nil}
```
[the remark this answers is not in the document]
SimeonPoot: I think the default configuration for the promUrl is "", right? so there should be no need to configure it here. It should then default to "", and this is not a valid URL to connect to prometheus.
I saw that the tests were green, so I did not really think about this twice...

kured-ds.yaml
```diff
@@ -30,8 +30,8 @@ spec:
       containers:
         - name: kured
           image: docker.io/weaveworks/kured
-                 # If you find yourself here wondering why there is no
-                 # :latest tag on Docker Hub,see the FAQ in the README
+                # If you find yourself here wondering why there is no
```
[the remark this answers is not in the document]
SimeonPoot: ok, my OCD was acting up... Fixed it like you said.

pkg/alerts/prometheus.go
```diff
@@ -7,42 +7,77 @@ import (
...
-func PrometheusActiveAlerts(prometheusURL string, filter *regexp.Regexp) ([]string, error) {
-	client, err := api.NewClient(api.Config{Address: prometheusURL})
+// PromClient is a wrapper around the Prometheus Client interface and implements the api
+// This way, the PromClient can be instantiated with the configuration the Client needs, and
+// the ability to use the methods the api has, like Query and so on.
+type PromClient struct {
+	papi papi.Client
+	api  v1.API
+}
+
```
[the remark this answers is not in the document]
SimeonPoot: yeah, I just dove into this while I was busy with this change.. so I'm not a 'know-it-all' either :) 
but, [https://github.com/prometheus/client_golang/blob/master/api/client.go](prometheus client_golang) using this gives us some defaults regarding the 'Roundtripper' (time-outs and so on).. so I think this is a good way to go. 

Also, I'm not sure if 'token expiration' was in scope before this change either, by using the prometheus client, we can for certainly have some more flexibilty regarding http traffic, like certificates, proxies and so on. 

The client going stale.. Is this a reported issue somewhere? I'm not sure how this could happen all of a sudden :)

pkg/alerts/prometheus.go
SimeonPoot: Looked into this again... You are absolutely right! I thought it was defaulting to RebootRequired.. But it isn't.
yes, let's add it to the README and remove it from here. As this alertName can be set to whatever the sheer imagination of the person configuring it.. :) 
This was done for the regex filter as well... I think this can be improved for the regex as well as the label filtering (in another fix), but for now it's not a show-stopper :)

pkg/alerts/prometheus.go
```diff
@@ -7,42 +7,77 @@ import (
 	"sort"
 	"time"
 
-	"github.com/prometheus/client_golang/api"
+	log "github.com/sirupsen/logrus"
+
+	papi "github.com/prometheus/client_golang/api"
 	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	"github.com/prometheus/common/model"
+	"k8s.io/apimachinery/pkg/api/errors"
```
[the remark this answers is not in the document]
SimeonPoot: No, fixed it.

cmd/kured/main.go
SimeonPoot: I do think both 'blockers' are doing in essence the same thing, that is.. but the implementation is different.. both are getting some input-filter, and returning a blocking-list... 
Yes, it's a bit confusing.. But I was trying to do the most 'pragmatic' thing to do, by opt-ing into existing functionality. 
That is; all alerts are being 'Queried', and then iterated over to get alertNames.. I merely added another 'label' check to each alert.
I think what makes it more readable (as stated into the method itself) would be to create separate functions (filterByRegex and filterByLabels) and let these functions do the specific filtering logic. 

I think adding another blocker, probably will lead into doing another query to the prometheus server...

SimeonPoot:
@evrardjp (and @jackfrancis) I'm going to split this PR... Because I don't think querying the labels has the correct logic. This could introduce a breaking change in the future. I will remove querying labels from the ActiveAlerts method, remove the Unit-tests that test this, and documentation. 
But will keep the structs, unit-tests for the regex to help improving the repo. 
And introduce another PR when querying the labels is better thought-through.

cmd/kured/main.go
```diff
@@ -531,7 +533,7 @@ func rebootAsRequired(nodeID string, rebootCommand []string, sentinelCommand []s
 
 		var blockCheckers []RebootBlocker
 		if prometheusURL != "" {
-			blockCheckers = append(blockCheckers, PrometheusBlockingChecker{promURL: prometheusURL, filter: alertFilter})
+			blockCheckers = append(blockCheckers, PrometheusBlockingChecker{filter: alertFilter, includeLabel: alertIncludeLabel})
```
jackfrancis, line 536: I don't think we want to entirely replace the existing prometheus regex blocking functionality, do we?

I would advocate that we preserve the existing functionality if the `--alert-filter-regexp` configuration is passed in, and make this new `--alert-include-label` configuration a mutually exclusive option.

So:

1) add CLI validation that only one configuration is possible (i.e., you aren't allowed to include `--alert-include-label` and `--alert-filter-regexp` at the same time
2) append the `blockCheckers` array above depending on which of the above prometheus configurations are being used

SimeonPoot: Hi @jackfrancis, thanks for your feedback. The first bulletpoint in the PR states that existing code should work... And it is working, or am I missing something perhaps? 
Why shouldn't we be allowed to include both filtering by regex and labels? 

Just to clarify, you want to have the prometheus-client configuration into the PrometheusBlockingChecker struct?

jackfrancis: Oh, I was confused by the removal of the `promURL` property. I see you discussed that with @evrardjp already. So ignore this comment. :)

cmd/kured/main.go
```diff
@@ -245,7 +245,9 @@ type KubernetesBlockingChecker struct {
 }
 
 func (pb PrometheusBlockingChecker) isBlocked() bool {
-	alertNames, err := alerts.PrometheusActiveAlerts(pb.promURL, pb.filter)
+	prom, _ := alerts.PromClient{}.New(papi.Config{Address: prometheusURL})
```
jackfrancis, line 248: Should we ignore the error here?

cmd/kured/main.go
SimeonPoot: Also, As I was testing/checking/investigating, I stumbled upon the Alert() function of the prometheus client.. I was wondering why are we using the Query() function of the prometheus client if there's also an Alert() function that goes to the /api/v1/alerts and gets all Active Alerts. This might give us an easier way to get alerts, and filter them accordingly.  
what do you guys think @evrardjp @jackfrancis

jackfrancis: Sounds like an improvement to me!

pkg/alerts/prometheus.go
SimeonPoot: removed it

cmd/kured/main_test.go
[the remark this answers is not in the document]
SimeonPoot: incorporated the client into the test with an broken url, the tests passed (You can see the error popping up if you go with `go test ./... -v`). 
tested this also on kind cluster.

cmd/kured/main.go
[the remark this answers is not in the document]
SimeonPoot: incorporated the promClient into PrometheusBlockingChecker

SimeonPoot:
@evrardjp and @jackfrancis  (Sorry to spam you guys!!) 
I reverted all query active alerts by labels, because I didn't find the logic holding. So this is just PR 1, of 2, but I need some time to figure out how to handle querying labels the correct way. (I've to check prometheus, kubernetes source code, to see how they're doing this). 
Also, perhaps I'll take a look at the Alert() function of the prometheus client, to see if this will come in handy.  
I left the Unit-tests, as well as the prometheus client configuration in the PR as I think this should give more quality/extensibility to the project.
Regards, Simeon Poot

kured-ds.yaml
```diff
@@ -15,7 +15,7 @@ spec:
     matchLabels:
       name: kured
   updateStrategy:
-   type: RollingUpdate
+    type: RollingUpdate
```
jackfrancis, line 18: while these whitespace fixes *are* readability improvements, let's omit them from this PR as they aren't related

pkg/alerts/prometheus.go
```diff
@@ -7,22 +7,40 @@ import (
...
+// This way, the PromClient can be instantiated with the configuration the Client needs, and
+// the ability to use the methods the api has, like Query and so on.
+type PromClient struct {
+	papi papi.Client
+	api  v1.API
+}
+
+// New creates a new client to the Prometheus API.
+// It returns an error on any problem.
+func (p PromClient) New(conf papi.Config) (*PromClient, error) {
```
jackfrancis, line 25: I think we want an exportable `New()` func which returns a new instance, rather than a method against the type struct that does so. So this instead:

```go
func New(conf papi.Config) (*PromClient, error) {
        promClient, err := papi.NewClient(conf)
 	if err != nil {
 		return nil, err
 	}
 	client := PromClient{papi: promClient, api: v1.NewAPI(promClient)}
 	return &client, nil
}
```

jackfrancis (changes requested):

kured-ds.yaml
SimeonPoot: allright

pkg/alerts/prometheus.go
SimeonPoot: I agree on this one. fixed it.

pkg/alerts/prometheus.go
```diff
@@ -7,22 +7,39 @@ import (
...
+// This way, the PromClient can be instantiated with the configuration the Client needs, and
+// the ability to use the methods the api has, like Query and so on.
+type PromClient struct {
+	papi papi.Client
+	api  v1.API
+}
+
+// New creates a new client to the Prometheus API.
+// It returns an error on any problem.
+func New(conf papi.Config) (*PromClient, error) {
```
jackfrancis, line 25: this is looking good, thanks!

I think reading through this, it would be clearer if we rename `New` to `NewPromClient`. Sorry for all the change requests, I think we're almost there.

jackfrancis (changes requested):

pkg/alerts/prometheus.go
SimeonPoot: I had the same feeling. I proceeded with it, because of the name of instantiation promClient. If it were a prometheus package, then it would read good enough, like prometheus.New(). 
I'll change it this evening. Thanks for the feedback and help.

jackfrancis (approved):
code lgtm

@evrardjp are you aware if the PR E2E tests that run have prometheus coverage?

jackfrancis (approved):
/lgtm

PR #394, opened 2021-06-17
Added rebootCommand values

renaudhager:
Hi all,

This PR adds `--reboot-command` flag in the chart.

Let me know if anything needs to be changed.

### DIFF
--- a/charts/kured/values.yaml
+++ b/charts/kured/values.yaml
@@ -31,6 +31,7 @@ configuration:
   prometheusUrl: ""          # Prometheus instance to probe for active alerts
   rebootDays: []             # only reboot on these days (default [su,mo,tu,we,th,fr,sa])
   rebootSentinel: ""         # path to file whose existence signals need to reboot (default "/var/run/reboot-required")
+  rebootCommand: "/sbin/systemctl reboot" #command to run when a reboot is required by the sentinel
   slackChannel: ""           # slack channel for reboot notfications
   slackHookUrl: ""           # slack hook URL for reboot notfications
   slackUsername: ""          # slack username for reboot notfications (default "kured")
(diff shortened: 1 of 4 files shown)


ckotzbauer:
Thanks for your contribution @renaudhager.
Could you please increase the `version` in the `Chart.yaml` file from `2.6.0` to `2.7.0` as its a feature-addition. This also should fix the failing Github-Check for testing.

renaudhager:
> Thanks for your contribution @renaudhager.
> Could you please increase the `version` in the `Chart.yaml` file from `2.6.0` to `2.7.0` as its a feature-addition. This also should fix the failing Github-Check for testing.

Done.

renaudhager:
Some checks seem unhappy, but I don't understand why. Let me know what I need to change to fix them.

charts/kured/values.yaml
```diff
@@ -31,6 +31,7 @@ configuration:
   prometheusUrl: ""          # Prometheus instance to probe for active alerts
   rebootDays: []             # only reboot on these days (default [su,mo,tu,we,th,fr,sa])
   rebootSentinel: ""         # path to file whose existence signals need to reboot (default "/var/run/reboot-required")
+  rebootCommand: "/sbin/systemctl reboot" #command to run when a reboot is required by the sentinel
```
ckotzbauer, line 34: This `# command` should to the trick instead of `#command`

ckotzbauer:
Could you please add the new param to the README of the chart? If all checks are green then, this should be ready 😉

renaudhager:
> Could you please add the new param to the README of the chart? If all checks are green then, this should be ready 😉

Done, hopefully it should all good this time :-)

ckotzbauer:
I'll have a look later today what's wrong here.

ckotzbauer (approved):

alexeldeib:
@renaudhager I noticed in this PR, the chart value is set differently from the in-code default used by Kured. Was that deliberate? It breaks the default chart deployment on platforms like AKS, ref https://github.com/weaveworks/kured/issues/404

jackfrancis:
@alexeldeib good find!

https://github.com/weaveworks/kured/pull/408

PR #402, opened 2021-07-04
link to versioned shoutrrr docs

piksel:
shoutrrr now have versioned docs to allow directly linking to the version that matches the one you use
changes should always backwards compatible, but not the other way around


dholbach (approved):
Nice one - thanks for the PR @piksel!

PR #410, opened 2021-07-23
Add --alert-firing-only parameter to only consider firing alerts

### LINKED ISSUE #393: Kured blocking reboot on pending alerts
I have been struggling with the auto rebooting nodes using Kured coz of pending alerts which is not an issue for us.
Since we have so many alerts it comes and goes from pending state randomly and blocks Kured most of the time.

Suggestion: It would be better to just block on "Firing Alerts" or an option to set if we want to block on "Firing Alert" or "Pending Alert" or both.

MattJeanes:
This change allows the user to specify `--alert-firing-only=true` to change the Prometheus blocking checker to only consider firing alerts instead of also pending alerts, which can be a common and normal occurrence in clusters.

Fixes #393

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ The following arguments can be passed to kured via the daemonset pod template:
 ```console
 Flags:
       --alert-filter-regexp regexp.Regexp   alert names to ignore when checking for active alerts
+      --alert-firing-only bool              only consider firing alerts when checking for active alerts
       --blocking-pod-selector stringArray   label selector identifying pods whose presence should prevent reboots
       --drain-grace-period int              time in seconds given to each pod to terminate gracefully, if negative, the default value specified in the pod will be used (default: -1)
       --skip-wait-for-delete-timeout int    when seconds is greater than zero, skip waiting for the pods whose deletion timestamp is older than N seconds while draining a node (default: 0)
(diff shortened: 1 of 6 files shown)


ckotzbauer (commented):
Thanks for your contribution. I only had a quick look at your changes, but I think it looks good in general. Could you please split the code-changes and the chart-changes to two separate PRs?

MattJeanes:
@ckotzbauer thank you for your review, I have removed the Helm chart changes from this PR and raised a new PR for them here: https://github.com/weaveworks/kured/pull/415

MattJeanes:
As for details on manual testing, I have been running these changes in 4 AKS clusters since Friday with no issues and tested with the flag in all 3 states (true, false, missing) successfully with both pending and firing alerts to ensure they function correctly.

An exact test I performed involved the new image with no changes to the command line arguments and a pending and firing alert. The app reported in logs that the both alerts were preventing a reboot. After adding the flag `--alert-firing-only=true` and restarting Kured it now dropped to only show the firing alert in the logs preventing a reboot. After clearing this firing alert (pending one remaining) the machine successfully rebooted.

I also wasn't sure looking at the [DEVELOPMENT.md](https://github.com/weaveworks/kured/blob/main/DEVELOPMENT.md) if I should increment versions in my PR or if that was down to the Kured team, but in the charts PR I incremented the chart version from `2.8.0` to `2.8.1` as the charts GitHub action was failing about this when it ran.

ckotzbauer (approved):
LGTM, but there's a rebase required now.
/cc @jackfrancis @evrardjp

MattJeanes:
Of course there are changes directly conflicting with mine 😄 I have rebased now and as a bonus have added a test!

evrardjp (approved):
LGTM

PR #415, opened 2021-07-26
Add --alert-firing-only parameter to chart

MattJeanes:
These are the appropriate chart changes for #410


charts/kured/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.7.0"
 description: A Helm chart for kured
 name: kured
-version: 2.8.0
+version: 2.8.1
```
ckotzbauer, line 5: Could you change this to 2.9.0, as its a feature-addition?

ckotzbauer (approved):
This looks very good.
**Info:** This should not be merged until #410 is included in a kured release!

charts/kured/Chart.yaml
MattJeanes: I've changed this to `2.9.0` now.

PR #418, opened 2021-08-03
Add `reboot-delay` command line argument

amorey:
Currently, kured issues the system reboot command immediately after
kubectl drain finishes.

This is a problem for processes that need extra time to finish but aren't
running on pods and therefore aren't controlled by kubectl drain (e.g.
de-registering nodes from external load balancers).

This patch solves the problem by introducing a `reboot-delay` command
line argument that can be used to add a delay after kubectl drain
finishes but before the reboot command is issued.

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -41,6 +41,7 @@ var (
 	// Command line flags
 	forceReboot                     bool
 	drainTimeout                    time.Duration
+	rebootDelay                     time.Duration
 	period                          time.Duration
 	drainGracePeriod                int
 	skipWaitForDeleteTimeoutSeconds int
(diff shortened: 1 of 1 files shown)


kingdonb:
(I released the e2e checks that were waiting for approval.) Thanks again!

evrardjp:
Thanks for splitting your PR in two! I will have a look and think of all the things good and bad about this. Eventually. (yes that's my sense of humour!)

evrardjp (approved):
I think this is fine.

evrardjp:
I realise now it's missing documentation and changes into helm chart + manifest.  Could you do a PR for that @amorey ?

PR #419, opened 2021-08-04
Add `label-with-exclude-from-external-lbs` CLI argument to enable graceful removal/addition from external load balancers

### LINKED ISSUE #358: EC2 instance graceful restart
Is there a recommended way to initiate a graceful restart of EC2 instances from kured? For my particular use case, I'd like to de-register instances from ELBs before rebooting the OS but I can't see a way to do this with the kured command line options.

amorey:
Previously, kured issued the system reboot command without first removing nodes from any connected external load balancers (ELBs).
    
This behavior caused downtime on restart because ELBs send traffic to kube-proxy pods running on nodes until the ELB health checks fail or the node is de-registered explicitly.
    
This patch solves the problem by adding a command line argument (`label-with-exclude-from-external-lbs`) that, when enabled, adds a "node.kubernetes.io/exclude-from-external-load-balancers" label to nodes undergoing kured reboot. This label tells the Kubernetes control plane to de-register the affected node from any connected ELBs. The node label is removed after restart which causes the control plane to re-register the node with the ELBs.

Close https://github.com/weaveworks/kured/issues/358

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -71,6 +71,7 @@ var (
 	rebootEnd     string
 	timezone      string
 	annotateNodes bool
+	labelWithELBX bool
 
 	// Metrics
 	rebootRequiredGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
@@ -87,6 +88,15 @@ const (
 	KuredRebootInProgressAnnotation string = "weave.works/kured-reboot-in-progress"
 	// KuredMostRecentRebootNeededAnnotation is the canonical string value for the kured most-recent-reboot-needed annotation
 	KuredMostRecentRebootNeededAnnotation string = "weave.works/kured-most-recent-reboot-needed"
+)
+
+const (
+	// ExcludeFromELBsLabelKey is a label key that tells the K8S control plane to exclude a node from external load balancers
+	ExcludeFromELBsLabelKey = "node.kubernetes.io/exclude-from-external-load-balancers"
+	// ExcludeFromELBsLabelVal is a label value used to track label placement by kured
+	ExcludeFromELBsLabelVal = "kured-remove-after-reboot"
+	// ExcludeFromELBsLabelKeyEscaped is the escaped label key value passed to the Patch() function
+	ExcludeFromELBsLabelKeyEscaped = "node.kubernetes.io~1exclude-from-external-load-balancers"
 )
 
 func init() {
@@ -164,6 +174,8 @@ func main() {
 
 	rootCmd.PersistentFlags().BoolVar(&annotateNodes, "annotate-nodes", false,
 		"if set, the annotations 'weave.works/kured-reboot-in-progress' and 'weave.works/kured-most-recent-reboot-needed' will be given to nodes undergoing kured reboots")
+	rootCmd.PersistentFlags().BoolVar(&labelWithELBX, "label-with-exclude-from-external-lbs", false,
+		"if set, the label 'node.kubernetes.io/exclude-from-external-load-balancers' will be added to nodes undergoing kured reboots")
 
 	rootCmd.PersistentFlags().StringVar(&logFormat, "log-format", "text",
 		"use text or json log format")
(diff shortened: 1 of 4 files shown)


evrardjp:
I will review asap.

evrardjp:
> > Little detail: I think this should be opt-in.
> 
> What would you suggest calling the command line argument?

I have no strong opinion :)
Maybe "--label-with-exclude-from-external-lb" ?

evrardjp:
I think it would be wise to document that future CLI flag's purpose in our documentation + include it (commented out) in our manifest.

amorey:
Are there any changes that need to be made to the PR?

ckotzbauer (approved):

cmd/kured/main.go
```diff
@@ -492,6 +552,9 @@ func rebootAsRequired(nodeID string, rebootCommand []string, sentinelCommand []s
 		log.Fatal(err)
 	}
 
+	// Remove ExcludeFromELBs label immediately to allow ELB registration
+	disableExcludeFromELBs(client, nodeID)
```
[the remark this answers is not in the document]
evrardjp: It does an api call, so it doesn't do nothing ;)

However, it made me realise that an error in disableExcludeFromELBs (e.g. one that the API can throw) is not used in any way... Why do we return the error if it's to do nothing with it? How should we behave if the API call fails?

evrardjp (commented):
I think I would like to discuss the guard and the last bit of the code.

cmd/kured/main.go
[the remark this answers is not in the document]
amorey: Currently the error is being used in the unit tests.

evrardjp:
I don't want to be a blocker for progress. If we feel it's a good feature, then let's move ahead... please let's just solve git conflicts and human ones ;)

cmd/kured/main.go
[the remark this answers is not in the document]
amorey: What would you like to do here? Do you want to guard the disable and make API errors fatal?

amorey:
> I don't want to be a blocker for progress. If we feel it's a good feature, then let's move ahead... please let's just solve git conflicts and human ones ;)

Sounds good to me! I rebased with main, fixed the conflicts and am waiting for feedback on the changes you requested above.

amorey:
https://github.com/weaveworks/kured/pull/419#pullrequestreview-764506186

I fenced in `disableExcludeFromELBs()` and modified the code to fail if the API returns an error. Is there anything left to do before this can be merged?

cmd/kured/main.go
```diff
@@ -493,6 +553,12 @@ func rebootAsRequired(nodeID string, rebootCommand []string, sentinelCommand []s
 		log.Fatal(err)
 	}
 
+	if labelWithELBX {
```
jackfrancis, line 556: Don't we want to move this remove node label logic down into the the part where we know we're holding the lock? Similar to where we do `if annotateNodes && !rebootRequired` and then delete node annotations?

amorey: The `disableExcludeFromELBs()` method can be run safely with or without the lock. Are there any particular cases you're thinking of where you would want the method to run only if the lock is present?

jackfrancis: I'm just thinking that if we do this when we're not holding the lock, that means we're doing it in every iteration. And that means that we're calling the apiserver for the node object every time. On a cluster with hundreds of thousands of nodes, with a small kured interval (e.g., 1 min), that means we're going to potentially be sending many apiserver requests every second.

If I understand correctly, the only reason we want to disable this label is to remove it after a successful reboot. In every scenario that fits that description, we *should* have the lock (because we only ever release the lock inside that block).

Does that make sense?

cc @dholbach @ckotzbauer for thoughts as well

ckotzbauer: I agree with @jackfrancis, that we should avoid requesting the apiserver to excessive. The logic should be safe when only called with a lock being present.

PR #439, opened 2021-09-28
[WIP] Add quay.io as second registry

ckotzbauer:
This PR adds `quay.io` as second container-registry to our build-scripts.
The secret-variables for the quay.io credentials are WIP.
Maybe we should mention this change in the docs?


dholbach (approved):
From what I've heard, the secrets should be in place now... :crossed_fingers: 

Thanks for the PR!

PR #442, opened 2021-09-28
Updated go binary in kured docker file

AnamikaKhantwal:


ckotzbauer:
Hi @AnamikaKhantwal, thanks for your PR.
Can you explain a bit why you opened this PR and why you introduced the `dc.Dockerfile` and `dc.Makefile`?

In the current main the Dockerfile is already updated to `alpine:3.14` which solves some vulnerabilities.
In the current main we also updated to Golang `1.16` to be aligned with the supported Kubernetes versions (1.20-1.22) from Kured. Therefor we updated all `k8s.io/*` golang packages to `0.21.x` to be compatible with the three mentioned Kubernetes versions regarding their supported minor-version-skew.
This all will be released in the upcoming 1.8.0 release.

AnamikaKhantwal:
> Hi @AnamikaKhantwal, thanks for your PR. Can you explain a bit why you opened this PR and why you introduced the `dc.Dockerfile` and `dc.Makefile`?
> 
> In the current main the Dockerfile is already updated to `alpine:3.14` which solves some vulnerabilities. In the current main we also updated to Golang `1.16` to be aligned with the supported Kubernetes versions (1.20-1.22) from Kured. Therefor we updated all `k8s.io/*` golang packages to `0.21.x` to be compatible with the three mentioned Kubernetes versions regarding their supported minor-version-skew. This all will be released in the upcoming 1.8.0 release.

Thanks @ckotzbauer for the explanation, closing my PR now.

PR #445, opened 2021-10-08
Revert "Add quay.io as second registry"

dholbach:
Reverts weaveworks/kured#439

:sob:


evrardjp:
Hello folks!

What was the reason for the revert? I fail to see that in the commit message.

dholbach:
Weaveworks's quay.io account is still not working. Maybe we should move to GHCR? I'll ask again. :-(

PR #446, opened 2021-10-08
Add ghcr.io as second registry

dholbach:
Please don't merge just yet.


dholbach:
Getting 1.8.2 out will be a good test-case. We can document the location of both registries (and how to use the Helm chart) in the main README afterwards?

Issue #227, opened 2020-11-10
blocking-pod-selector blocks on non-running pods


roy-work:
`kured`'s `--blocking-pod-selector` appears to only look for the existence of `Pod`s matching the given selector. This means that pods in, e.g., the state `Completed` (such as a finished `CronJob`) will block reboots; I don't believe this is how this feature is intended to work.

We run `kured` in a Kubernetes cluster, and we have a few `CronJob`s that we run that we want to block reboots. Thus, we configured `kured` with `--blocking-pod-selector=kured-prevent-reboot=true`; we then set that label on the relevant `CronJob`s. We are seeing `kured` refuse to reboot nodes, even when those jobs aren't running.

The code I think is involved in this process is [here](https://github.com/weaveworks/kured/blob/64202ff440ae51e5fe3eafb0c614426f638bc53f/cmd/kured/main.go#L174-L196), and appears pretty straight-forward: it merely lists out pods according to the provided label, and adds a field selector to restrict to only pods scheduled on the same VM as the `kured` that is doing the query. I think the only thing it misses is that it isn't ignoring pods in the `Completed` or `Error` states; they _exist_, but they're not really running, and preventing reboots means a `CronJob` that — even though it doesn't overlap `kured`'s window — keeps getting scheduled on the same node can prevent reboots indefinitely by just having the `Completed` pod hanging around.

---

As an example, In our logs, we see:

```
time="2020-11-08T00:59:52Z" level=warning msg="Reboot blocked: matching pods: [AAA-1604739600-v876j BBB-1604732400-pr2ps CCC-1602165600-fptr4 CCC-1602165600-j8m4f DDD-1604296800-qx5pb]"
```
(I've masked the names of the jobs.)

As an example, the `-v876j` `Pod` is first created:
```
"I1107 09:00:08.288281    2228 kubelet.go:1901] SyncLoop (ADD, \"api\"): \"AAA-1604739600-v876j_eee(<snip>)\""
```
About 10 minute later, it finishes:
```
"I1107 09:11:01.820871    2228 kubelet.go:1946] SyncLoop (PLEG): \"AAA-1604739600-v876j_eee(<snip>)\", event: &pleg.PodLifecycleEvent{ID:\"<snip>\", Type:\"ContainerDied\", Data:\"<snip>\"}"
```
Then, _much_ later, we get the above message from `kured`:
```
[...]

PR #459, opened 2021-11-05
Exclude terminated pods from the blocking mechanism

georgekaz:
Terminated pods should be excluded from the blocking a reboot as per https://github.com/weaveworks/kured/issues/227

This adds status filters to the fieldSelector in order to do that. I've not updated tests here but have successfully tested the exact same filter using kubectl

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -284,7 +284,7 @@ func (pb PrometheusBlockingChecker) isBlocked() bool {
 }
 
 func (kb KubernetesBlockingChecker) isBlocked() bool {
-	fieldSelector := fmt.Sprintf("spec.nodeName=%s", kb.nodename)
+	fieldSelector := fmt.Sprintf("spec.nodeName=%s,status.phase!=Succeeded,status.phase!=Failed,status.phase!=Unknown", kb.nodename)
 	for _, labelSelector := range kb.filter {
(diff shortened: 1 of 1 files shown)


evrardjp (commented):
I am not against that. Maybe it can be tested in a lab, see how it behaves? It will tough to test it in CI.... Maybe @dholbach  can try in his test cluster? :D

georgekaz:
I've been running this code in my cluster for the past 3 weeks. Things have rebooted as expected but I've not deliberately created blocking situations. Thanks

georgekaz:
Actually, I correct myself, I have seen this working properly in prod. We had a badly configured cronjob that was stuck in the `Pending` phase and it correctly blocked the node from restarting, whereas all the nodes that had completed cronjobs rebooted fine.

The relevant blocking selector in this case is the label `job-name`:
```
          blockingPodSelector: [     # label selector identifying pods whose presence should prevent reboots
            "block-reboots=true",
            "job-name"
          ]
```

evrardjp (approved):
Not tested, but I don't see it as a reason to block this :)

georgekaz:
Thanks @evrardjp . How do we go about getting this merged? Will someone pick it up?

ckotzbauer (approved):

PR #460, opened 2021-11-16
WIP windows support for kured

marosset:
Signed-off-by: Mark Rossetti <marosset@microsoft.com>

Here is a prototype that I have working for Windows support.

A few things to note include:
- This requires WindowsHostProcessContainers feature (introduced in v1.22) to be enabled for kube-apiserver and kubelet
  - This also means that kubeclient/API refs **must** be v1.22 or greater
  - WindowsHostProcessContainers is on-track for beta in v1.23
- InClusterConfig does not work with HostProcessContainers today (see not and linked issue in code)
- An init container is used to register a scheduled task that runs on startup to delete `c:\var\run\reboot-required` if it exists. This way things outside of the Windows servicing stack can write the sentinel file and expect a reboot to be triggered _once_.
- The script to check for pending reboots will write `c:\var\run\reboot-required` if one of several registry keys controlled by the servicing stack is set. These keys are cleared as servicing operations are processed.
- Windows escapes `"`'s in command line args (these interfere with command construction if not trimmed)
- This PR does not address how the Windows container image should be built. I left some open questions / comments in the dockerfile / makefile for discussion tho


cmd/kured/main.go
```diff
@@ -199,13 +205,19 @@ func newCommand(name string, arg ...string) *exec.Cmd {
...
-func buildHostCommand(pid int, command []string) []string {
+func buildHostCommand(pid int, command []string, GOOS string) []string {
 
-	// From the container, we nsenter into the proper PID to run the hostCommand.
-	// For this, kured daemonset need to be configured with hostPID:true and privileged:true
-	cmd := []string{"/usr/bin/nsenter", fmt.Sprintf("-m/proc/%d/ns/mnt", pid), "--"}
-	cmd = append(cmd, command...)
-	return cmd
+	if GOOS == windows {
+		// On Windows kubed daemonset needs to run in a HostProcess container which
```
jackfrancis, line 211: nit: s/kubed/kured

scripts/Test-PendingReboot.ps1
```diff
@@ -0,0 +1,23 @@
+<#
+    Return 0 / SUCCESS if reboot is required!
+#>
+
+$cbsKey = Get-ItemProperty "HKLM:\Software\Microsoft\Windows\CurrentVersion\Component Based Servicing" -ErrorAction Ignore
+if ($cbsKey.PSObject.Properties.name -contains 'RebootPending') {
+    Write-Output "Reboot requried - Component Based Servicing:RebootPending registry value detected"
+    New-Item -Path '/var/run' -Name 'reboot-required' -ItemType File -Force
```
jackfrancis, line 8: Remind me: why do we want to emulate the sentinel filepath pattern here, why not just return 0 (reboot required) if any of the two keys here have the required reboot properties?

marosset: Sure - Both of those registry keys are only cleared on the next boot if Windows servicing operations set those keys.
If a user manually sets either of those keys and then reboots this would most likely result in the node being stuck in a reboot loop.
Adding a sentinel file here seemed like a more generic approach. I can image situations where things outside of the windows servicing stack would want to trigger a reboot once (things like node-problem-detector for example).
I also feel it is important to be able to detect when there is a pending reboot due to things like Windows Update needing to reboot the node for security fixes to be applied. Having a powershell script that can detect both situations seemed a good compromise.

jackfrancis: That makes sense. So basically the idea here is to promote the general idea of "create the sentinel file `c:\var\run\reboot-required` to indicate that you want a reboot.

And the init container that we're shipping is a POC for how to implement an end-to-end solution (because you need a thing to delete that file at bootup to make this strategy work).

cmd/kured/main.go
```diff
@@ -93,6 +96,10 @@ func init() {
 }
 
 func main() {
+	os.Stdout.WriteString(fmt.Sprintf("Command line args: %v\n", os.Args))
```
evrardjp, line 99: do we need this line?  Do we need it in main?

cmd/kured/main.go
```diff
@@ -93,6 +96,10 @@ func init() {
...
+	os.Stderr.WriteString(fmt.Sprintf("Command line args: %v\n", os.Args))
+	log.Warnf("Command line args: %v\n", os.Args)
```
evrardjp, line 101: Do we need to warn?

cmd/kured/main.go
```diff
@@ -93,6 +96,10 @@ func init() {
 }
 
 func main() {
+	os.Stdout.WriteString(fmt.Sprintf("Command line args: %v\n", os.Args))
+	os.Stderr.WriteString(fmt.Sprintf("Command line args: %v\n", os.Args))
```
evrardjp, line 100: Is that really worth sending into stderr?

cmd/kured/main.go
```diff
@@ -199,13 +205,19 @@ func newCommand(name string, arg ...string) *exec.Cmd {
...
-	// From the container, we nsenter into the proper PID to run the hostCommand.
-	// For this, kured daemonset need to be configured with hostPID:true and privileged:true
-	cmd := []string{"/usr/bin/nsenter", fmt.Sprintf("-m/proc/%d/ns/mnt", pid), "--"}
-	cmd = append(cmd, command...)
-	return cmd
+	if GOOS == windows {
+		// On Windows kubed daemonset needs to run in a HostProcess container which
+		// run all processes in the hosts process namespace.
+		return command
+	} else {
```
evrardjp, line 214: Shouldn't we start to build an interface instead, and make sure the linker/user input allows us to infer the type of command to run?

cmd/kured/main.go
```diff
@@ -483,9 +495,24 @@ func deleteNodeAnnotation(client *kubernetes.Clientset, nodeID, key string) {
...
 func rebootAsRequired(nodeID string, rebootCommand []string, sentinelCommand []string, window *timewindow.TimeWindow, TTL time.Duration, releaseDelay time.Duration) {
-	config, err := rest.InClusterConfig()
-	if err != nil {
-		log.Fatal(err)
+
+	log.Infof("In rebootAsRequired")
+	var config *rest.Config
+	var err error
+
+	if runtime.GOOS == windows {
```
evrardjp, line 503: If we implement interface with windows/linux k8s servers as types, then we could have a method for getting the config based on those. It would look simpler and simpler to test.

cmd/kured/main.go
```diff
@@ -601,7 +628,33 @@ func rebootAsRequired(nodeID string, rebootCommand []string, sentinelCommand []s
 
 // buildSentinelCommand creates the shell command line which will need wrapping to escape
 // the container boundaries
-func buildSentinelCommand(rebootSentinelFile string, rebootSentinelCommand string) []string {
+func buildSentinelCommand(rebootSentinelFile string, rebootSentinelCommand string, GOOS string) []string {
```
evrardjp, line 631: see comment above.

evrardjp (changes requested):
I didn't fully review, but I feel like we can improve the code readability by implementing interfaces and the right types.

cmd/kured/main.go
marosset: No, sorry these were to help me figure out an issue related to https://github.com/microsoft/hcsshim/pull/1207.
I'll remove all of these.

evrardjp: Well, you found it useful to have flags displayed, so we should probably address that in a later patch :)

cmd/kured/main.go
marosset: removed, that was for debugging :)

cmd/kured/main.go
marosset: I'll look into building an interface to for windows/linux behaviors so we can remove the GOOS if/else statements in the code.
I wanted to cleanup some other parts of this PR first (like building the container images with make targets) but that is done.

cmd/kured/main.go
```diff
@@ -37,6 +35,7 @@ import (
 
 var (
 	version = "unreleased"
+	windows = "windows"
```
marosset, line 38: I'll remove this

cmd/kured/os_specific_behaviors.go
```diff
@@ -0,0 +1,93 @@
...
+	"fmt"
+	"os"
+
+	"github.com/google/shlex"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+type osSpecificBehaviors interface {
```
marosset, line 13: I'm open to suggestions for better names for any of the objects in this file :-/

scripts/Test-PendingReboot.ps1
```diff
@@ -0,0 +1,23 @@
...
+}
+
+$autoUpdateKey = Get-ItemProperty "HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\WindowsUpdate\Auto Update" -ErrorAction Ignore
+if ($autoUpdateKey.PSObject.Properties.name -contains 'RebootRequired') {
+    Write-Output "Reboot required - Auto Update:RebootRequired registry value detected"
```
rchaganti, line 13: Write-Output returns objects to the pipeline. We do not use it in a script unless that behavior is expected. Instead, you can use Write-Information.

PS C:\scripts> .\Test-PendingReboot.ps1 
Reboot not required

scripts/Test-PendingReboot.ps1
```diff
@@ -0,0 +1,23 @@
...
+    New-Item -Path '/var/run' -Name 'reboot-required' -ItemType File -Force
+}
+
+if (Test-Path -Path '/var/run/reboot-required') {
+    Write-Output "Reboot required"
+    exit 0
```
rchaganti, line 19: if your intention is to return 0 or 1, you must use `return` keyword and not `exit`. exit kills the PowerShell session where this function runs.

```
PS C:\scripts> $return = .\Test-PendingReboot.ps1 

PS C:\scripts> $return
Reboot not required
```
If you see in this, the return value from the script is not 0 or 1. Instead, it returned Reboot not required. If you replace the Write-Output with Write-Information and exit with return, here is what you will see.

```
PS C:\scripts> $return = .\Test-PendingReboot.ps1 

PS C:\scripts> $return
1
```

rchaganti (commented):
Adding a couple of comments around Test-PendingReboot.ps1 to ensure the code written follows the PowerShell best practices.

ckotzbauer:
@marosset Are there any news on this? Is this still WIP?

PR #464, opened 2021-11-25
bind environment variables to cobra flags with viper

### LINKED ISSUE #383: Add ability to set cli flags with environment variables
Hey folks 👋🏼 

Would be great if we could define the CLI flags also with Environment Variables, this is especially important for arguments that are somewhat sensitive (like `--notify-url`) where we mutate the value from Vault.

Thanks!

dkulchinsky:
fixes #383

Environment Variables should be prefixed with `KURED_` and match the command flag names (`-` replaced with `_`)

if values are specified through both the command flag and environment variable, command flag takes precedence.

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -37,11 +37,6 @@ import (
 	"github.com/weaveworks/kured/pkg/timewindow"
 )
 
-const (
-	// The environment variable prefix of all environment variables bound to our command line flags.
-	envPrefix = "KURED"
-)
-
 var (
 	version = "unreleased"
 
@@ -95,6 +90,8 @@ const (
 	KuredRebootInProgressAnnotation string = "weave.works/kured-reboot-in-progress"
 	// KuredMostRecentRebootNeededAnnotation is the canonical string value for the kured most-recent-reboot-needed annotation
 	KuredMostRecentRebootNeededAnnotation string = "weave.works/kured-most-recent-reboot-needed"
+	// EnvPrefix The environment variable prefix of all environment variables bound to our command line flags.
+	EnvPrefix = "KURED"
 )
 
 func init() {
@@ -207,7 +204,7 @@ func flagCheck(cmd *cobra.Command, args []string) {
 func bindViper(cmd *cobra.Command, args []string) error {
 	v := viper.New()
 
-	v.SetEnvPrefix(envPrefix)
+	v.SetEnvPrefix(EnvPrefix)
 	v.AutomaticEnv()
 	bindFlags(cmd, v)
 
@@ -220,7 +217,7 @@ func bindFlags(cmd *cobra.Command, v *viper.Viper) {
 		// Environment variables can't have dashes in them, so bind them to their equivalent keys with underscores
 		if strings.Contains(f.Name, "-") {
 			envVarSuffix := strings.ToUpper(strings.ReplaceAll(f.Name, "-", "_"))
-			v.BindEnv(f.Name, fmt.Sprintf("%s_%s", envPrefix, envVarSuffix))
+			v.BindEnv(f.Name, fmt.Sprintf("%s_%s", EnvPrefix, envVarSuffix))
 		}
 
 		// Apply the viper config value to the flag when the flag is not set and viper has a value
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -99,6 +99,15 @@ func init() {
 }
 
 func main() {
+	cmd := NewRootCommand()
+
+	if err := cmd.Execute(); err != nil {
+		log.Fatal(err)
+	}
+}
+
+// NewRootCommand construct the Cobra root command
+func NewRootCommand() *cobra.Command {
 	rootCmd := &cobra.Command{
 		Use:               "kured",
 		Short:             "Kubernetes Reboot Daemon",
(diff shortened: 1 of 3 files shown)


cmd/kured/main.go
```diff
@@ -95,11 +103,14 @@ func init() {
...
-		Run:    root}
-
+		Use:               "kured",
+		Short:             "Kubernetes Reboot Daemon",
+		PreRun:            flagCheck,
+		PersistentPreRunE: bindViper,
+		Run:               root}
+
+	rootCmd.PersistentFlags().StringVar(&nodeID, "node-id", "",
+		"node name kured runs on, should be passed down from spec.nodeName via KURED_NODE_ID environmen variable.")
```
ckotzbauer, line 113: typo in `environmen`

cmd/kured/main.go
```diff
@@ -182,16 +193,44 @@ func flagCheck(cmd *cobra.Command, args []string) {
 		log.Warnf("Deprecated flag(s). Please use --notify-url flag instead.")
 		trataURL, err := url.Parse(slackHookURL)
 		if err != nil {
-			log.Warnf("slack-hook-url is not properly formatted...no notification will be sent: %v\n", err)
+			log.Warnf("slack-hook-url is not properly formatted... no notification will be sent: %v\n", err)
 		}
 		if len(strings.Split(strings.Trim(trataURL.Path, "/services/"), "/")) != 3 {
-			log.Warnf("slack-hook-url is not properly formatted...no notification will be sent: %v\n", err)
+			log.Warnf("slack-hook-url is not properly formatted... no notification will be sent: unexpected number of / in URL\n")
```
ckotzbauer, line 199: I'm not sure if this should be changed in this PR. Maybe open another one to change the message.

ckotzbauer (commented):
Thanks for the PR. It looks really good, but I haven't had time to test it yet.
I've made two small comments.

cmd/kured/main.go
```diff
@@ -95,11 +103,14 @@ func init() {
...
 	rootCmd := &cobra.Command{
-		Use:    "kured",
-		Short:  "Kubernetes Reboot Daemon",
-		PreRun: flagCheck,
-		Run:    root}
-
+		Use:               "kured",
+		Short:             "Kubernetes Reboot Daemon",
+		PreRun:            flagCheck,
+		PersistentPreRunE: bindViper,
```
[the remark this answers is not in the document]
dkulchinsky: @evrardjp addressed the first comment (switched the order)

about Viper/Cobra binding - by far not the expert, I was mostly following the blog here https://carolynvanslyck.com/blog/2020/08/sting-of-the-viper/ and relevant Viper docs (https://github.com/spf13/viper#working-with-environment-variables)

1. To my understanding, that is the intended method, at this point it binds to the define cobra commands and just performs the environment variable binding if relevant

2. bindViper defines the prefix of env variables to consider `KURED_*` and `AutomaticEnv` does the matching between flags & env variables, with flags taking precedence and respecting their defaults, same validation for the values is applied as currently used for the flag values (for example if you use `KURED_LOCK_TTL=10` <-- that won't work, since it's not a `Time.duration`, the `bindFlags` function handles normalizing the matching of env variables to flags by replacing `-` with `_` - this just augments `AutomaticEnv` to handle the flags that have `-` in them that `AutomaticEnv` doesn't handle on it's own.

cmd/kured/main.go
dkulchinsky: fixed

cmd/kured/main.go
dkulchinsky: reverted

cmd/kured/main.go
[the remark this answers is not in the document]
dkulchinsky: As for testing, I tried looking into adding a unit test, however since Viper will only populate the flags from environment variable when the command is executed I don't see a (straightforward) way to make it work, since it fails to run due to missing serviceaccount token file.

perhaps this can be added to the kind tests, but I'm a bit out of my depth there...

OTOH, since KURED_NODE_ID now gets populated using this new code, this is actually being tested already :-)

dkulchinsky:
@ckotzbauer @evrardjp any additional thoughts? comments? 😄

jackfrancis:
@dkulchinsky does this change the name of the environment variables expected as input, or just add an additional `KURED_`-prefixed env var to every pre-existing variable?

If we change the name of the environment variables, that's a significant breaking change for existing users, and we'd normally only want to make that kind of change with a new major release (e.g., 2.0).

dkulchinsky:
> @dkulchinsky does this change the name of the environment variables expected as input, or just add an additional `KURED_`-prefixed env var to every pre-existing variable?
> 
> If we change the name of the environment variables, that's a significant breaking change for existing users, and we'd normally only want to make that kind of change with a new major release (e.g., 2.0).

Hey @jackfrancis, kured doesn't handle environment variables today at all (only command line flags via Cobra), the only exception was `KURED_NODE_ID` (which was read directly via `os.Getenv` here: https://github.com/weaveworks/kured/blob/ff1a27ba8b2c9bebc7f986f8a84cf24640e63dca/cmd/kured/main.go#L632), and it remains unchanged since it already had a `KURED_` prefix (only will be handled through Viper).

This PR doesn't introduce any breaking changes, environment variables are optional and command line flags still take precedence.

jackfrancis (approved):
lgtm

ckotzbauer (approved):

dkulchinsky:
Thank you folks!

PR #481, opened 2021-12-11
[WIP] ARM-Support

ckotzbauer:
This is a WIP-PR of a build with ARM support.
I'm not really sure about the changes in the Makefile. With the changes for windows-support in mind from #460 this would create more and more complicated tooling-scripts to maintain.

My personal favourite is to do only the go-compilation and other small tasks in the makefile and move the image build and -publish to Github Actions which are already able to create multiarch-builds out of the box. But this would remove the ability to create images locally (without writing the commands on your own).

I also would suggest to not separate different architectures by image-tag and use the same tags for different arches.

WDYT?

Ref: #23

### DIFF
--- a/.github/workflows/on-main-push.yaml
+++ b/.github/workflows/on-main-push.yaml
@@ -38,8 +38,10 @@ jobs:
 
       - name: Build image
         run: |
-          make DH_ORG="${{ github.repository_owner }}" image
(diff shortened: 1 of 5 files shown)


ckotzbauer:
superseded by #496

PR #494, opened 2022-01-11
Add ability to define ds annotations in helm chart

weseven:
Hi,
 this PR adds the ability to define custom annotations for the kured daemonset through the `dsAnnotations` helm value.

Value name and template syntax have been chosen to be similar to the already existing `podAnnotations`.

This covers our kured use case where during kured first setup all reboots are temporarily suspended by locking the daemonset using the annotation on the daemonset (ref. https://github.com/weaveworks/kured#disabling-reboots).

### DIFF
--- a/charts/kured/Chart.yaml
+++ b/charts/kured/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.9.1"
 description: A Helm chart for kured
 name: kured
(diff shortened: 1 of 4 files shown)


ckotzbauer (approved):

ckotzbauer:
Thanks @weseven for your PR. The new version is already released! 🎉

PR #496, opened 2022-01-12
Added support for multi-arch image build

davideshay:
Modified workflow actions to use QEMU and docker buildx to create multi-arch image and manifest in docker hub and ghcr.
Required a multi-arch compatible Dockerfile as well.


davideshay:
Might also close #23 .  arm can easily be added by just adding to the platform list, as would a few other archs.

davideshay:
These changes have been added. I was trying to add caching to further optimize. I would normally have the cache refer to the "latest" build, which would pickup, for example, the existing 1.91 image when building 1.92 automatically.  But this project doesn't want to publish a "latest" build. 

We could add some other docker tag when building from main, like "main-cache". Or you could just use a "latest" tag and let the user decide for themselves if they want to use it or not, and continue to recommend not using it for pulls.

Thoughts?

davideshay:
Can this be merged? And then if further build speed optimizations are needed, we can deal with them separately?

ckotzbauer:
This needs a bit more discussion in our next dev-meeting and more reviews to get merged. It is always possible to improve things in follow-up PRs.

cmd/kured/Dockerfile.multi
```diff
@@ -0,0 +1,21 @@
...
+ENV GOOS=$TARGETOS
+ENV GOARCH=$TARGETARCH
+ENV GOVARIANT=$TARGETVARIANT
+
+WORKDIR /src
+COPY go.mod go.sum .
+RUN go mod download
+COPY . .
+RUN go list -f '{{join .Deps "\n"}}' ./cmd/kured | grep -v /vendor/ | xargs go list -f '{{if not .Standard}}{{ $dep := . }}{{range .GoFiles}}{{$dep.Dir}}/{{.}} {{end}}{{end}}'
+RUN CGO_ENABLED=0 go build -o cmd/kured/kured cmd/kured/*.go
```
ckotzbauer, line 16: @davideshay Can we use a makefile-target here instead for compiling the binaries? This would avoid the duplication of the godeps-stuff. Also the `main.Version` is missing here.

ckotzbauer (changes requested):

cmd/kured/Dockerfile.multi
davideshay: The problem with running it via Makefile is that you don't get the advantages of the single run of buildx to auto-generate everything in the platform list, as well as create the multi-arch manifest.  There certainly could be a way to get the benefits of both, but I'm not aware right now of how.

On the main version piece, my thinking was that on-main-push would execute everytime there was a push to main, and that workflow run would create an image like "main-0fe532a1" with the short commit tag. When you create a new tag, that would run on-tag workflow, which I modified to use the version tag, publishing main:1.9.1

ckotzbauer: I didn't mean to change the whole architecture from this PR with buildx, that's all fine. My suggestion is only to create a make target similar to `make cmd/kured/kured` which has no hardcoded GO* env-args regarding the OS and ARCH. Then the godeps line from the Dockerfile can be reused in the makefile. (So only replace L15+16 from the Dockerfile)

ckotzbauer:
I tested this PR with my home-cluster (mixed amd64 and arm64) and it worked like charm.

cmd/kured/Dockerfile.multi
ckotzbauer: I think this PR would be ready to merge, when this can be addressed. @davideshay

ckotzbauer: @davideshay kindly ping on this.
If you don't have enough time right now, I think we can do this in a follow-up PR too.

davideshay:
I think this is ready to go now -- sorry couldn't quite figure out how to squash the commits on this. But I have now added the build command to come from the Makefile.  Turns out I don't really need the DEPS logic there because it's a new environment every build, but I still left the go build command in the Makefile, and changed the Dockerfile to call the new target.

ckotzbauer:
@davideshay Thanks for the change!
Do we really need the new manual workflow?

davideshay:
> @davideshay Thanks for the change! Do we really need the new manual workflow?

I found it was very difficult to test without it, but really not needed. Do you want me to delete it?

davideshay:
Done.

ckotzbauer (approved):

davideshay:
Not needed, was very difficult for me to test without it though.

PR #541, opened 2022-05-11
Add node labelling

### LINKED ISSUE #509: Support node readiness labels
The Zalando PostgreSQL Operator depends on node readiness labels to schedule pods on nodes.  To trigger a failover of the "leader" role in the PostgreSQL Patroni HA cluster from one pod to another in the event of a node reboot, a label must be updated on the node before it is cordoned and drained. The label must be updated once more after the node is rebooted and uncordened.

Would you accept a PR to optionally update node readiness labels before and after the reboot process?

liger1978:
Fixes #509

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,28 +3,30 @@
 
 <img src="https://github.com/weaveworks/kured/raw/main/img/logo.png" align="right"/>
 
-- [Introduction](#introduction)
-- [Kubernetes & OS Compatibility](#kubernetes--os-compatibility)
-- [Installation](#installation)
-- [Configuration](#configuration)
-  - [Reboot Sentinel File & Period](#reboot-sentinel-file--period)
-  - [Reboot Sentinel Command](#reboot-sentinel-command)
-  - [Setting a schedule](#setting-a-schedule)
-  - [Blocking Reboots via Alerts](#blocking-reboots-via-alerts)
-  - [Blocking Reboots via Pods](#blocking-reboots-via-pods)
-  - [Prometheus Metrics](#prometheus-metrics)
-  - [Notifications](#notifications)
-  - [Overriding Lock Configuration](#overriding-lock-configuration)
-- [Operation](#operation)
-  - [Testing](#testing)
-  - [Disabling Reboots](#disabling-reboots)
-  - [Manual Unlock](#manual-unlock)
-  - [Automatic Unlock](#automatic-unlock)
-  - [Delaying Lock Release](#delaying-lock-release)
-- [Building](#building)
-- [Frequently Asked/Anticipated Questions](#frequently-askedanticipated-questions)
-  - [Why is there no `latest` tag on Docker Hub?](#why-is-there-no-latest-tag-on-docker-hub)
-- [Getting Help](#getting-help)
+- [kured - Kubernetes Reboot Daemon](#kured---kubernetes-reboot-daemon)
+  - [Introduction](#introduction)
+  - [Kubernetes & OS Compatibility](#kubernetes--os-compatibility)
+  - [Installation](#installation)
+  - [Configuration](#configuration)
+    - [Reboot Sentinel File & Period](#reboot-sentinel-file--period)
+    - [Reboot Sentinel Command](#reboot-sentinel-command)
+    - [Setting a schedule](#setting-a-schedule)
+    - [Blocking Reboots via Alerts](#blocking-reboots-via-alerts)
+    - [Blocking Reboots via Pods](#blocking-reboots-via-pods)
+    - [Adding node labels before and after reboots](#adding-node-labels-before-and-after-reboots)
+    - [Prometheus Metrics](#prometheus-metrics)
+    - [Notifications](#notifications)
+    - [Overriding Lock Configuration](#overriding-lock-configuration)
+  - [Operation](#operation)
+    - [Testing](#testing)
+    - [Disabling Reboots](#disabling-reboots)
+    - [Manual Unlock](#manual-unlock)
+    - [Automatic Unlock](#automatic-unlock)
+    - [Delaying Lock Release](#delaying-lock-release)
+  - [Building](#building)
+  - [Frequently Asked/Anticipated Questions](#frequently-askedanticipated-questions)
+    - [Why is there no `latest` tag on Docker Hub?](#why-is-there-no-latest-tag-on-docker-hub)
+  - [Getting Help](#getting-help)
 
 ## Introduction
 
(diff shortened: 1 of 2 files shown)


ckotzbauer (commented):
Hi @liger1978,
thanks for your PR and working on the issue. This is much appreciated!

Some thoughts to the label-feature:
- The labels are only added, never removed. The zalando-usecase is toggling the node-label-value. Would it be better to remove the `before-labels` when the `after-labels` are applied? I think this is more in line with user expectations.
- I'm not sure about the cli-flag-names. What about `pre-reboot-node-labels` and `post-reboot-node-labels`?

Can you please separate the spelling corrections and readme updates from this PR? It's hard to review, when there are many other changes which aren't directly related to a PR. Please  always open separate PRs for that.

Thanks!

/cc @evrardjp @jackfrancis Thoughts?

liger1978:
@ckotzbauer I'm not sure what "removing a label" means in this context. In ready state, nodes will already have a label like `zalando` with a value like `ready`. Before reboot this must be set to something else, like `notready` (causing the operator to move the database leader role to a pod on another node).  After reboot the label will be set back to `ready`.  I've tested this and it works with the Zalando feature. I'm not sure where you are suggesting a label should be removed in this workflow.

liger1978:
@ckotzbauer I've updated the PR with the other requested changes.

ckotzbauer:
Thanks for your changes.
What you want (and the zalando-usecase is) is, that you toggle values of the same node-label. However the implementation only adds two sets of labels.
From the wording, I'm not sure if all our users understand that this feature is only really useful if you add pre- and post-labels for the same label-keys.
There are no enforcements that all label-keys from `pre-reboot-node-labels` are also configured as `post-reboot-node-labels`. If this is something a component in the cluster (e.g. monitoring) expects, than the reboot is done and the label is still present and indicates a reboot.

liger1978:
@ckotzbauer I've added the verification and updated the readme as requested.

cmd/kured/main.go
```diff
@@ -205,6 +214,19 @@ func flagCheck(cmd *cobra.Command, args []string) {
...
+	var afterRebootNodeLabelKeys []string
+	for _, label := range preRebootNodeLabels {
+		preRebootNodeLabelKeys = append(preRebootNodeLabelKeys, strings.Split(label, "=")[0])
+	}
+	for _, label := range afterRebootNodeLabels {
+		afterRebootNodeLabelKeys = append(afterRebootNodeLabelKeys, strings.Split(label, "=")[0])
+	}
+	sort.Strings(preRebootNodeLabelKeys)
+	sort.Strings(afterRebootNodeLabelKeys)
+	if !reflect.DeepEqual(preRebootNodeLabelKeys, afterRebootNodeLabelKeys) {
```
jackfrancis, line 227: So the idea is that you want common label keys for both pre-cordon and post-cordon, but in practice the *values* of the labels will be different pre and post. Is that correct?

cmd/kured/main.go
```diff
@@ -205,6 +214,19 @@ func flagCheck(cmd *cobra.Command, args []string) {
 			notifyURL = fmt.Sprintf("slack://%s", strings.Trim(trataURL.Path, "/services/"))
 		}
 	}
+	var preRebootNodeLabelKeys []string
```
jackfrancis, line 217: We can combine these two variable declarations into a single line, e.g.:

```
var preRebootNodeLabelKeys, afterRebootNodeLabelKeys []string
```

cmd/kured/main.go
liger1978: @jackfrancis Correct. @ckotzbauer asked that I validate that the labels match.

cmd/kured/main.go
liger1978: OK, will do.

cmd/kured/main.go
```diff
@@ -534,6 +562,31 @@ func deleteNodeAnnotation(client *kubernetes.Clientset, nodeID, key string) erro
 	return nil
 }
 
+func addNodeLabels(client *kubernetes.Clientset, node *v1.Node, labels []string) {
```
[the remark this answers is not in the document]
liger1978: OK, will do.

liger1978:
@jackfrancis I have implemented your suggestions.

cmd/kured/main.go
```diff
@@ -534,6 +561,31 @@ func deleteNodeAnnotation(client *kubernetes.Clientset, nodeID, key string) erro
...
+
+	_, err = client.CoreV1().Nodes().Patch(context.TODO(), node.GetName(), types.StrategicMergePatchType, bytes, metav1.PatchOptions{})
+	if err != nil {
+		var labelsErr string
+		for _, label := range labels {
+			k := strings.Split(label, "=")[0]
+			v := strings.Split(label, "=")[1]
+			labelsErr += fmt.Sprintf("%s=%s ", k, v)
+		}
+		log.Fatalf("Error updating node labels %s via k8s API: %v", labelsErr, err)
```
ckotzbauer, line 585: Please use `Errorf` instead, we don't want to shutdown kured 😉

README.md
```diff
@@ -218,6 +220,17 @@ running job or a known temperamental pod on a node will stop it rebooting.
...
+### Adding node labels before and after reboots
+
+If you need to add node labels before and after the reboot process, you can use `--pre-reboot-node-labels` and `--after-reboot-node-labels`:
+
+```console
+      --pre-reboot-node-labels=zalando=notready
+      --after-reboot-node-labels=zalando=ready
+```
+
+Note that label keys specified by these two flags should match. If they do not match, a warning will be generated.
```
ckotzbauer, line 232: Please add a note, that the labels could be comma-delimited or the flags could be added multiple times.

cmd/kured/main.go
liger1978: Done

README.md
liger1978: Done.

ckotzbauer (approved):
From my perspective this looks good now. @jackfrancis do you want to have a final look?
I would include this into 1.10.0 if there are not objections.

jackfrancis (approved):
lgtm

liger1978:
Thanks. Please can it be merged!

erwbgy:
@liger1978 One minor niggle.  The current implementation has **pre**-reboot-node-labels and **after**-reboot-node-labels, which I find confusing.  IMO the latter should be **post**-reboot-node-labels as per the original suggestion from @ckotzbauer. 

Or alternatively **before**-reboot-node-labels and **after**-reboot-node-labels-reboot-node-labels.

ckotzbauer:
Yes, you are right @erwbgy. Thanks for the hint. This should be fixed.

liger1978:
@erwbgy @ckotzbauer @jackfrancis I have made this change.  Please review/approve/merge. Thank you!

ckotzbauer (approved):

liger1978:
@ckotzbauer Thank you very much for merging! Do you have any idea when we can expect a new release with these changes included?

ckotzbauer:
I think in the next week.

Issue #359, opened 2021-04-16
Redesign idea: "Kured Large scale"


evrardjp:
Hello everyone.

I guess this was supposed to be a blog post, but I will make it an issue instead. We can discuss it better here.

## Intro 
Our lovely kured works well for many. It's simple, and I would like to keep it that way.

Yet, we recently had more and more special cases, which requires building a bit more complexity around our code.
Slowly, we'll increase our complexity, and reduce our readability.

Next to this, we have seen PRs and issues in the past mentioning shortcomings of our current architecture, for flexibility [1][2] or security reasons [3].

I would like us to think about what "another kured" (for more complex cases) could be.
Here are my thoughts...

## Separate kured daemonset in components

Instead of having a single daemonset doing everything, I would separate in multiple components:
- An operator **brain**, analysing the cluster state, and orchestrating the **Node Reboot Procedure**(s)
- A **Node Reboot Detector**, a user-defined daemonset, which will probably hold one or multiple containers.

### The Brain

The brain would be based on the current code from `rebootAsRequired`. 
Instead of locking on the kured daemonset, it would lock on a separate object, like a config map.
The rest would be like the current model: checking for any reboot blockers, draining the node (which requires patch node permissions), and triggering a reboot of one to multiple nodes (depending on the **"Nodes Maintenance Window"**). Because the reboot wouldn't be local anymore, the brain would create a *job* to trigger a reboot scheduled on the necessary node, job which can be described in the **"Reboot Procedure"**.

Configuration of the behaviour of the brain would be done by editing custom resources and/or config maps.

### The "Node reboot detector"

This daemonset will detect if the node needs rebooting, and simultaneously report it back to the "brain" and optionally expose that metric to prometheus. The latter is similar to our `maintainRebootRequiredMetric`. We can have a consistent period check between the *brain* and the *detector* by making sure both of them read `period` from the same config map/CR.
Alternatively to "report it back to the brain", the brain could `fetch` the node information exposed with the `rebootRequiredMetric`.

The daemonset is likely to be a single container, which will check the presence of a file at an expected location `/kured/reboot-required`. This will only require read access to that volume. No need for `privileged` or `hostPIDs`.

For the basic ubuntu workloads, this is the same as having our daemonset, with a volume mapping `/var/run` to `/kured/` in read-only mode.

[...]

PR #566, opened 2022-05-27
 added fixed reboot time for single node

atighineanu:
This PR implements the possibility to
  schedule a reboot (one time, for one
  single node) from the config yaml.


atighineanu:
This is a hackish way to schedule 1 node for reboot at a fixed time. It led me to the idea we would need this[](https://github.com/weaveworks/kured/issues/359) soon.

PR #569, opened 2022-05-29
Downgrade kind-images for 1.22.x and 1.23.x

ckotzbauer:
Signed-off-by: Christian Kotzbauer <git@ckotzbauer.de>

Possible changes since 1.9.2 that came to my mind when analyzing these errors:
- Kind-Update from 0.11.0 to 0.14.0
- `k8s.io/*` updates from 0.22.4 to 0.23.6
- Retry-Cordon-PR

The problem seems to be the kind-image updates for 1.22.x and 1.23.x. With the newer versions (1.22.9, 1.23.6) the tests may fail, because the last node does not need a reboot. With the previous versions (1.22.4, 1.23.0) the problem does not occur on multiple runs. The new 1.24.0 image seems to be fine.


jackfrancis (approved):
lgtm

PR #589, opened 2022-06-25
Updated helm charts with new message template flag

atighineanu:
Updated helm charts with new message template flag
+ bumped chart version

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ Flags:
       --lock-annotation string              annotation in which to record locking node (default "weave.works/kured-node-lock")
       --lock-release-delay duration         hold lock after reboot by this duration (default: 0, disabled)
       --lock-ttl duration                   expire lock annotation after this duration (default: 0, disabled)
+      --message-template-uncordon string    message template used to notify about a node being successfully uncordoned (default "Node %s rebooted & uncordoned successfully!")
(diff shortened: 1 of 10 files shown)


ckotzbauer (changes requested):
Can you add the flag to the daemonset and the chart Readme too? Otherwise the change has no effect.

atighineanu:
> Can you add the flag to the daemonset and the chart Readme too? Otherwise the change has no effect.

Done.

ckotzbauer (approved):

jackfrancis (approved):
/lgtm

atighineanu:
Somethin else I need to do here? @ckotzbauer @jackfrancis @dholbach

ckotzbauer:
@jackfrancis We have to release a Kured patch-release now, the chart points to the current Kured release which doesn't have the new feature. Helm-PRs which needs Code-Changes can only be merged when there was a new Kured release.

jackfrancis:
@ckotzbauer thanks for the clarification

going forward let's add a /hold gesture (or hold tag) to indicate that a PR shouldn't be merged, hopefully I won't do this again! :/

ckotzbauer:
that's a good idea 👍 
I can try to cut a release in the evening.

PR #595, opened 2022-07-03
Integrate GoReleaser, Cosign and Syft

ckotzbauer:
Signed-off-by: Christian Kotzbauer <git@ckotzbauer.de>

This PR changes our build-system and introduces SBOM-Generation and image-signing.
There's a new make-target `bootstrap-tools` to prepare the local dev environment.

The multi-arch images are signed by cosign with ephemeral keys from Github OIDC tokens. The signature is pushed to the registry alongside the image.
A SBOM is generated by Syft for each image in the spdx format. The SBOM is also signed and pushed to the registries alongside the images and signatures.


evrardjp:
Very nice concept, thanks @ckotzbauer !

.github/scripts/goreleaser-install.sh
```diff
@@ -0,0 +1,37 @@
...
+}
+
+test -z "$TMPDIR" && TMPDIR="$(mktemp -d)"
+TAR_FILE="$TMPDIR/${FILE_BASENAME}_$(uname -s)_$(uname -m).tar.gz"
+export TAR_FILE
+
+(
+    echo "Downloading GoReleaser $VERSION..."
+    curl -sfLo "$TAR_FILE" \
+        "$RELEASES_URL/download/$VERSION/${FILE_BASENAME}_$(uname -s)_$(uname -m).tar.gz"
```
evrardjp, line 19: If we are really doing a supply chain security, shouldn't we statically provide this, so that we are always sure it's matching what we expect? Else, someone can just hijack the $RELEASES_URL with a matching version file, and put their own malware?

.github/workflows/on-main-push.yaml
```diff
@@ -40,20 +44,47 @@ jobs:
...
-          file: cmd/kured/Dockerfile.multi
           platforms: linux/arm64, linux/amd64, linux/arm/v7, linux/arm/v6, linux/386
           push: true
           tags: |
             docker.io/${{ GITHUB.REPOSITORY }}:main-${{ steps.tags.outputs.sha_short }}
             ghcr.io/${{ GITHUB.REPOSITORY }}:main-${{ steps.tags.outputs.sha_short }}
+
+      - name: Generate SBOM
+        run: |
+          .tmp/syft docker.io/${{ GITHUB.REPOSITORY }}:main-${{ steps.tags.outputs.sha_short }} -o spdx | jq --compact-output > kured-docker.sbom
```
evrardjp, line 74: If we do this after the move to a new org, does it still make sense to have the docker.io command?

evrardjp (commented):
I like the general cleanup done here. I need to dig deeper before giving my vote.

.github/workflows/on-main-push.yaml
ckotzbauer: fixed

.github/scripts/goreleaser-install.sh
ckotzbauer: If someone is able to hijack our scripts (in git or inside the actions runner), he/she could also change a hardcoded url. When we want to be safe against an attack on goreleaser's side, we need a hermetic build or several more checks in this scripts. That's not what I want to achieve here (even when I'm not yet fine with this personally).

This PR is only meant to be a very first step. The "Supply-Chain Levels for Software Artifacts" ([SLSA](https://slsa.dev/)) describes four levels. With this PR we don't even make the first one. So there will be follow ups.

dholbach (approved):
Great work @ckotzbauer

Issue #604, opened 2022-07-21
`metrics.labels` in chart seems to be incorrectly implemented


kingdonb:
The problem is that kube-prometheus-stack expects `ServiceMonitor` to be labeled a certain way, and when kube-prometheus-stack is installed with a release name such as `kube-prometheus-stack` the standard behavior is to create a prometheus with the label selector `release: kube-prometheus-stack`

This means that our serviceMonitor must also have this label, to be selected, and most charts call this `metrics.labels` as `additionalLabels`

However the Helm chart itself according to best practices affixes a `release: ...` label to everything

https://github.com/weaveworks/kured/blob/67c50b27ab8674280255d180e307f6fa70c65ccf/charts/kured/templates/servicemonitor.yaml#L10-L13

So here, I get an error like:

```
Helm upgrade failed: error while running post render on files: map[string]interface {}(nil): yaml: unmarshal errors:
```

the text of the error indicates the `release` label was already present, cannot be added again or overridden as Flux's Helm Operator does not permit messy repeat application and shadowing of labels (each field must be specified exactly once.)

So I understood this issue now, at least just well enough to write it down, but I don't know how it should be solved. I just hacked the ServiceMonitor manually for now, (I was inspired to start monitoring kured today as I just experienced a surprise reboot after upgrading, almost definitely attributable to kured, and I have only recently learned enough about prometheus and servicemonitors to start slanging them around properly just last week, thanks to having stumbled onto this article that explained the problem I was having conceptualizing how prometheus decides what to ingest and how to run down a trouble there from every angle: https://managedkube.com/prometheus/operator/servicemonitor/troubleshooting/2019/11/07/prometheus-operator-servicemonitor-troubleshooting.html)

kingdonb:
I went looking for an example that does this better and podinfo has almost the same thing where labels are applied, but the template differs in that it does not define a `release:` label there:

https://github.com/kingdonb/podinfo/blob/8f712e5e4f1daee0b1b73a25b14a706826a2bc44/charts/podinfo/templates/_helpers.tpl#L33-L43

It seems like that might be the way to fix it – simply don't put a `release` label on everything, I was mistaken about which labels are recommended as the Helm best practice:

https://helm.sh/docs/chart_best_practices/labels/#standard-labels

`release` is not anywhere in there, and although it does not say anywhere here that you cannot use the `release` label, I wonder why it was added, as it does seem to conflict with the default behavior of kube-prometheus-stack.

[...]

PR #605, opened 2022-07-21
Switch to standard labels from Helm best practices

kingdonb:
https://helm.sh/docs/chart_best_practices/labels/#standard-labels

Suggesting this as a fix for #604

### DIFF
--- a/charts/kured/README.md
+++ b/charts/kured/README.md
@@ -21,8 +21,31 @@ $ helm delete my-release
 
 The command removes all the Kubernetes components associated with the chart and deletes the release.
 
+## Upgrade Notes
+
+### From 2.x to 3.x
+
+The Helm chart labels have been realigned to conform with the [standard labels](https://helm.sh/docs/chart_best_practices/labels/#standard-labels) in the current Helm Chart Best Practices guide, so this upgrade will fail unless the DaemonSet is deleted and recreated. The only way that Helm supports delete and recreate is by uninstalling, so please uninstall the Kured Helm chart before installing again with `v3.x`.
+
+If you use any GitOps tool, please check and understand how to do a reinstall beforehand.
+
+Supposing users want to enable metrics and use a `ServiceMonitor` with the `kube-prometheus-stack` chart's default `prometheus` instance. Starting with a chart that has values:
+
+```
+metrics:
+  create: true
+  labels:
+    release: kube-prometheus-stack
+```
+
+A "ServiceMonitor" needs a "release" label to be discovered by the Prometheus-Operator with the default configuration of `kube-prometheus-stack` and this chart (in the prior `v2.x` chart) already sets a `release` label hardcoded. This is changed by applying the best-practise labels in the chart `v3.x`. Now the user can decide which `release` label-value should be used.
+
+With this update, it's more readily possible to make use of the Kured chart with `kube-prometheus-stack`'s default `ServiceMonitor` selector configuration.
 
 ## Migrate from stable Helm-Chart
+
+### From 1.x to 2.x
+
 The following changes have been made compared to the stable chart:
 - **[BREAKING CHANGE]** The `autolock` feature was removed. Use `configuration.startTime` and `configuration.endTime` instead.
 - Role inconsistencies have been fixed (allowed verbs for modifying the `DaemonSet`, apiGroup of `PodSecurityPolicy`)
--- a/charts/kured/Chart.yaml
+++ b/charts/kured/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: "1.10.1"
 description: A Helm chart for kured
 name: kured
-version: 2.17.0
+version: 3.0.0
 home: https://github.com/weaveworks/kured
 maintainers:
   - name: ckotzbauer
--- a/charts/kured/templates/_helpers.tpl
+++ b/charts/kured/templates/_helpers.tpl
@@ -57,16 +57,16 @@ Return the appropriate apiVersion for podsecuritypolicy.
 Returns a set of labels applied to each resource.
 */}}
 {{- define "kured.labels" -}}
-app: {{ template "kured.name" . }}
-chart: {{ template "kured.chart" . }}
-release: {{ .Release.Name }}
-heritage: {{ .Release.Service }}
+app.kubernetes.io/name: {{ template "kured.name" . }}
+helm.sh/chart: {{ template "kured.chart" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end -}}
 
 {{/*
 Returns a set of matchLabels applied.
 */}}
 {{- define "kured.matchLabels" -}}
-app: {{ template "kured.name" . }}
-release: {{ .Release.Name }}
+app.kubernetes.io/name: {{ template "kured.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end -}}


kingdonb:
I have tested it out here and it's working for me:

https://github.com/kingdonb/csh-flux/blob/494960cfb1a78bad3334158fc26241e3f0c95174/clusters/home-workers/kured/kured-helmrelease.yaml#L12-L23

It is definitely a breaking change, during the upgrade I got the error that daemonsets could not have a change in labels as it was forbidden by their immutability rules. If this issue was already documented somewhere else then I missed it.

ckotzbauer (changes requested):
Can you please bump the "version" field in the Chart.yaml to the next major version, add a short description to the chart-README how to migrate from 2.x to 3.x and ensure that the tests are green?
Thanks!

kingdonb:
Absolutely.

I have just haphazardly deleted the `DaemonSet` to get around the issue, since it is not possible to change labels on the daemonset and labels are immutable.

I will go back and test whether the `helm upgrade --force` works with this, and likewise if Helm Controller users can expect using this to help or not:

```
$ k explain helmrelease.spec.upgrade.force
KIND:     HelmRelease
VERSION:  helm.toolkit.fluxcd.io/v2beta1

FIELD:    force <boolean>

DESCRIPTION:
     Force forces resource updates through a replacement strategy.
```

I am honestly not sure if this works, since I have read recently about how replacement strategy does not always honor the same semantics as delete-and-recreate. So we may actually have to advise the users to delete and then recreate their DaemonSets through a Helm upgrade following the deletion.

Either way, I will test it a few times and make sure what I'm adding in the documentation reads well and is correct. 👍 This may take a couple of iterations though, please bear with...

kingdonb:
This is my PR but I suppose I'm going to encourage us to think about it for a little while before merging this, as a major version bump is no trifling matter, there may be other changes, and I can maintain this change in my own fork as long as I'm the only person who has needed it.

I would much rather bump the MAJOR chart version once we are certain there are no other breaking changes required, and so we can merge them all together at once – philosophically I think that it is better to bump the major version less in the entire lifetime of a project. I am not the maintainer on this project though, so if you feel stronger about this change than I do, feel free to merge and release. 👍

It will make my life easier not having to maintain a separate fork in the long run, but FWIW that's a sacrifice I'm more than willing to make for the moment

kingdonb:
I think I've said everything I thought that needed to be explained in the `charts/kured/README.md` now 👍

ckotzbauer:
Thanks for putting so much thought into the upgrade process and breaking change. I like that very much. :+1: 

A major version jump with big changes is not as bad or rare for helm charts as it is for the software itself. (Kured itself does not get a new major version).
Many charts get a major update every few months (or sometimes even every few weeks). These label-changes are special, because it is really one of the few situations where a complete reinstall is necessary. We need to make users aware of this, correct. That's why it has to be in the chart readme. (See https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#upgrading-chart as an example, each major release is briefly described here). But since Kured does not have a state, a reinstallation is very easy. With a GitOps tool (e.g. Flux2), the existing HelmRelease has to be deleted first and then reinserted with a second commit with the new version.

kingdonb:
I squashed all my changes here, I will add one more commit that highlights first that uninstall and reinstall will work.

Then I'm sure I can trim some of the text that followed. I'd like to keep some explanation about the immutable labels just so that folks know why this uninstall/reinstall is necessary, but we probably don't need the whole story about Helm Controller.

kingdonb (commented):
👍 LGTM

charts/kured/README.md
```diff
@@ -21,8 +21,40 @@ $ helm delete my-release
...
+
+The `kube-prometheus-stack` chart, by default, expects any `ServiceMonitor` to have a label `release` which matches the `kube-prometheus-stack` installation's `Release.Name`. Kured was already setting the same `release` label on every YAML resource to its own Helm chart `Release.Name` until this Helm chart v3.x, which made the default configuration for Kured metrics incompatible with the default for `ServiceMonitor`.
+
+Labels on a DaemonSet are immutable so `helm upgrade --force` will also fail:
+
+```
+Error: cannot patch "kured" with kind DaemonSet: DaemonSet.apps "kured" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"release":"kube-prometheus-stack"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
+```
+
+Users do not need to make any configuration change, only to uninstall and reinstall the `kured` chart. Manually deleting the DaemonSet and recreating it via the next upgrade will also work in environments where direct cluster access is permitted.
```
ckotzbauer, line 52: (Z 46-52): This can be removed, as this is the Kubernetes behaviour which can't be changed. Maybe one sentence can be added to the previous text-block which shortly descibes, that changing labels on existing objects needs a recreation.

ckotzbauer (commented):
Thanks for your efforts. I think this can be shortened on several places. This has much todo with default behaviours of Helm and Kubernetes (which should not be descibed in detail here) and the fact that each cluster operator has to know how he/she has to handle their infrastructure tools, which should also not be part of the Kured documentation.

charts/kured/README.md
kingdonb: It's not that changing labels on existing objects needs a recreation though, it's DaemonSet in particular which is immutable. (Anyone who uses Kured must use it in DaemonSet mode though, so the fact is probably not really necessary to mention.)

charts/kured/README.md
```diff
@@ -21,8 +21,40 @@ $ helm delete my-release
...
 
+## Upgrade Notes
+
+### From 2.x to 3.x
+
+The Helm chart labels have been realigned to conform with the [standard labels](https://helm.sh/docs/chart_best_practices/labels/#standard-labels) in the current Helm Chart Best Practices guide, so this upgrade will fail unless the DaemonSet is deleted and recreated. The only way that Helm supports delete and recreate is by uninstalling and reinstalling.
+
+With a GitOps tool (e.g. Flux), the existing HelmRelease has to be deleted first and pruned, then apply the HelmRelease with a second commit and the new version.
+
+\*Unfortunately, [Helm does not support delete and recreate](https://github.com/fluxcd/flux2/discussions/2854#discussioncomment-3057748) when using upgrades, making upgrades requiring an immutable field change impossible.
```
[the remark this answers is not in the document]
kingdonb: 853d476

charts/kured/README.md
ckotzbauer: Label-Selectors are immutable (e.g. in the PodSpec of the DaemonSet)

charts/kured/README.md
```diff
@@ -21,8 +21,40 @@ $ helm delete my-release
...
+
+```
+metrics:
+  create: true
+  labels:
+    release: kube-prometheus-stack
+updateStrategy: RollingUpdate
+```
+
+The `kube-prometheus-stack` chart, by default, expects any `ServiceMonitor` to have a label `release` which matches the `kube-prometheus-stack` installation's `Release.Name`. Kured was already setting the same `release` label on every YAML resource to its own Helm chart `Release.Name` until this Helm chart v3.x, which made the default configuration for Kured metrics incompatible with the default for `ServiceMonitor`.
```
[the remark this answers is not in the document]
kingdonb: 14afbef890721ea135b3b128b3f10ce95ec46ace

charts/kured/README.md
kingdonb: e42317eca41bf8f273be3faa46f349218f399ac7

kingdonb: If there is a finer point in the language that I've gotten subtly wrong then it's beyond my understanding, welcome to suggest any corrections or further deletions if you prefer. I'll accept them.

charts/kured/README.md
```diff
@@ -21,8 +21,36 @@ $ helm delete my-release
...
+  labels:
+    release: kube-prometheus-stack
+updateStrategy: RollingUpdate
+```
+
+A "ServiceMonitor" needs a "release" label to be discovered by the Prometheus-Operator with the default configuration of `kube-prometheus-stack` and this chart already sets a `release` label hardcoded. This is changed by applying the best-practise labels in the chart `v3.x`. Now the user can decide which `release` label-value should be used.
+
+This made the use of the Kured chart metrics incompatible with the defaults for `kube-prometheus-stack`'s `ServiceMonitor`.
+
+Labels on a DaemonSet are immutable so `helm upgrade --force` will fail too.
```
kingdonb, line 46:
```suggestion
Label selectors on a DaemonSet are immutable so `helm upgrade --force` will fail too.
```

kingdonb: Is this the suggested change 👍

kingdonb:
Noodling around with my releases, I think I see what you're saying, the concern about labels and labelselectors is not only for the daemonset; after my experimentation, I wound up with surprising combinations of labels in places I did not expect them and the end result is Prometheus selected my serviceMonitor, but the service failed to match!

Maybe better if the directions say more forcefully that only the uninstall and reinstall approach is supported.

charts/kured/README.md
```diff
@@ -21,8 +21,32 @@ $ helm delete my-release
...
+If you use any GitOps tool, please check and understand how to do a reinstall beforehand.
+
+Supposing users want to enable metrics and use a `ServiceMonitor` with the `kube-prometheus-stack` chart's default `prometheus` instance. Starting with a chart that has values:
+
+```
+metrics:
+  create: true
+  labels:
+    release: kube-prometheus-stack
+updateStrategy: RollingUpdate
```
ckotzbauer, line 39: Can you remove the "updateStrategy" here? It is unrelated to the labels.

ckotzbauer (commented):
Thanks for your changes so far, one more comment.

kingdonb:
I will squash the fixup commits once more, LMK if there's anything else you notice 👍

ckotzbauer (approved):

ckotzbauer:
@jackfrancis @evrardjp @dholbach Does anybody want to take another look since it's a major chart change?

jackfrancis (approved):
lgtm

this will be a bit painful but there's nothing imminent _forcing_ users to upgrade to 3.x

kingdonb:
The fortunate thing is that users should not need to do any configuration change, an uninstall-reinstall is totally sufficient. And also fortunate that there are no stateful components, so for any users who are doing GitOps, this reinstall should be very easy (as is pinning the chart version to `^2.0.0`.) 👍 

I'm not sure what that failure is, but I'm guessing it is a flaky test that can be ignored. It did not show up in any earlier builds.

PR #628, opened 2022-09-08
Add reboot in progress prometheus metric

Aaron-ML:
Signed-off-by: Aaron Lucas [github@alucas.me](mailto:github@alucas.me)

Added a new metric to give a heads up when a node is in the process of rebooting. 

We occasionally run into alerting issues during reboot when daemonsets or pods crashloop and eventually recover.

Adding this gives us the ability to run inhibits on these crashloops during the reboot window.

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -86,6 +86,12 @@ var (
 		Name:      "reboot_required",
 		Help:      "OS requires reboot due to software updates.",
 	}, []string{"node"})
+
+	rebootInProgressGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: "kured",
+		Name:      "reboot_in_progress",
+		Help:      "Node in progress of rebooting for software updates.",
+	}, []string{"node"})
 )
 
 const (
@@ -101,6 +107,7 @@ const (
 
 func init() {
 	prometheus.MustRegister(rebootRequiredGauge)
+	prometheus.MustRegister(rebootInProgressGauge)
 }
 
 func main() {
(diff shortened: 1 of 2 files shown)


cmd/kured/main.go
```diff
@@ -531,6 +538,31 @@ func maintainRebootRequiredMetric(nodeID string, sentinelCommand []string) {
 	}
 }
 
+func maintainRestartInProgressMetric(nodeID string) {
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		log.Fatal(err)
```
ckotzbauer, line 544: Please use `Error` instead. Non-functional metrics should not kill kured.

ckotzbauer (commented):
Hi @Aaron-ML, thanks for your PR.
When I read the code, the only thing the new metric does is, that it represents the existance of the DaemonSet-Lock in conjunction with the nodeID.

Did you instead tried to write a Prometheus-Rule which has the Daemonset-Annotations joined? The nodeID is there also available. I think this could do the trick.

cmd/kured/main.go
Aaron-ML: I originally triggered off of reboots but smaller nodes reboot so fast that the metric cannot get scraped before kured reboots. I figured a node lock was somewhat safe to indicate a reboot in progress was going to occur.

ckotzbauer:
But, this new metric is not "faster" to recognize than the existing metrics from kube-state-metrics unless you set a very low scrape-interval for kured. I'm not sure if this is the way to go.
The problem with monitoring is, that it is always delayed and situations which dissappear very fast can't handled correctly very often.

Aaron-ML:
> But, this new metric is not "faster" to recognize than the existing metrics from kube-state-metrics unless you set a very low scrape-interval for kured. I'm not sure if this is the way to go. The problem with monitoring is, that it is always delayed and situations which dissappear very fast can't handled correctly very often.

Certainly possible to export this in kube-state-metrics however it will either be null or 1 and working with labels versus metrics you generally lose a lot of flexibility. 

I understand that it's possible to miss this metric in scraping, I figured setting the metric at lock time would give enough headroom to scrape within reason.

Looking back on it now an empty node still might miss that window since the lock to reboot time is pretty low, however I think decreasing scrape interval is pretty low overhead unless I'm mistaken. Happy to go either way with this.

ckotzbauer:
@jackfrancis @evrardjp What are your thoughts on this?

Issue #519, opened 2022-03-29
Release-Build does not use newest Golang version available


ckotzbauer:
I just published 1.9.2 of Kured which uses go 1.16. The `actions/setup-go` Action resolves `1.16` to `1.16.14` which is not the latest patch-release. This would be `1.16.15`. #511 is not resolved with our new release.
Maybe we have to think about pinning the Golang version in the builds to a specific version.
WDYT?

ckotzbauer:
@evrardjp  @jackfrancis

PR #632, opened 2022-09-14
Update golang and x/net to avoid CVE in our builds

evrardjp:


evrardjp:
Interestingly, the gh action doesn't take latest 1.18, and takes 1.18.5...

ckotzbauer:
@evrardjp See #519, I discovered this problem in the past already. The fix seems not to work then :disappointed:

evrardjp:
> @evrardjp See #519, I discovered this problem in the past already. The fix seems not to work then disappointed

Commit number 2 seems to fix the problem, and uses the latest 1.18. 
Is that what you meant?

evrardjp:
For clarity's sake, I don't think the CVE really impacts us strongly, but it's worth updating anyway.

jackfrancis (approved):
lgtm

ckotzbauer (approved):

PR #638, opened 2022-09-19
require DCO from now on

dholbach:

### DIFF
--- a/DCO
+++ b/DCO
@@ -1,0 +1,36 @@
+Developer Certificate of Origin
+Version 1.1
+
+Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
+660 York Street, Suite 102,
+San Francisco, CA 94110 USA
+
+Everyone is permitted to copy and distribute verbatim copies of this
+license document, but changing it is not allowed.
+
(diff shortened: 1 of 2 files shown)


evrardjp (commented):
Two questions.
1) Would it make sense to also add the DCO bot, or do we wait for this to be in the next organisation? 
2) I thought the DCO needed to apply for the CNCF org, which means this PR should land after the move. WDYT?

dholbach:
I added the DCO bot to the `kubereboot` earlier organisation. As the repo transfer is requested now, not sure if it makes sense to enable it to `weaveworks/kured` for the last couple of commits.

dholbach:
I think PRs will need to be re-targeted after the org move, so maybe let's just merge it, as it's just a docs change for now.

evrardjp:
The PRs will be moved over, so I don't think it's a big issue.

evrardjp:
If you don't mind, I will merge https://github.com/weaveworks/kured/pull/637 before this one. Else, the CI has to run again for a doc change.

evrardjp (approved):

PR #645, opened 2022-09-20
Remove charts from repo

evrardjp:
As we'll be extracting charts to the new repository, there is no need to keep the current helm charts in this repository.

This allows us to have a cleaner code.


dholbach (approved):
Thanks @evrardjp

dan-leanix:
@dholbach @evrardjp Might it be that this change breaks Helm chart deployments? We received today on our Terraform runs the following error:

```
Error: looks like "https://weaveworks.github.io/kured" is not a valid chart repository or cannot be reached: failed to fetch https://weaveworks.github.io/kured/index.yaml : 404 Not Found
```

dholbach:
@dan-leanix Can you update to `kubereboot.github.io/charts` and make sure you have 4.0.1?

dan-leanix:
@dholbach https://github.com/kubereboot/kured/issues/654 I am on it. Looking at https://kubereboot.github.io/charts/index.yaml and the URLs `3.0.0` should still work.

dan-leanix:
@dholbach `kubereboot.github.io/charts` fixed the issue for us. Thanks again for the quick response.

PR #649, opened 2022-09-20
Deprecate bot account

dholbach:


dholbach:
Let's go with #650 instead.

evrardjp:
Thanks Daniel, sorry I made a similar contribution at the same time...

PR #650, opened 2022-09-20
Use github token instead of bot accounts

evrardjp:
This should be one credential less to manage.


dholbach:
Well done adding the labels, etc.!

Issue #308, opened 2021-02-17
Move off of peter-evans/link-checker GH action


dholbach:
Move off of peter-evans/link-checker GH action, it's been replaced with https://github.com/lycheeverse/lychee-action

jackfrancis:
It looks like lycheeverse/lychee-action doesn't support recursive usages (traversing into all child directories), see:

https://github.com/lycheeverse/lychee/issues/78

Since both kured workflows that use peter-evans/link-checker use its `-r` option (recursively search), should we wait until lycheeverse/lychee-action supports the recursive before making this change?

dholbach:
Looks like https://github.com/lycheeverse/lychee/pull/165 is close, so we can wait until a fix is released. :-)

evrardjp:
Update: Work not finished in lychee 165, maybe someone can help there? : )

PR #677, opened 2022-10-26
Use lychee link checker and add Fossa badge

dholbach:
Move to lychee action as well.

Fixes: #308


evrardjp (approved):
For the link checker:
This should cover enough files. The previous concern of recursive shouldn't be a big deal.
On top of that, most of the content should be eventually moved to the documentation website.

For the license check, nothing to say, looks like a good addition.

Issue #642, opened 2022-09-20
Adopt governance guidelines for Kured


dholbach:
One of the items on https://github.com/cncf/toc/issues/922 is 

> Is your project working on written, open governance? see https://contribute.cncf.io/maintainers/governance/

A simple piece of governance rules for our project could be https://github.com/cncf/project-template/blob/main/GOVERNANCE-maintainer.md which was written by the Governance WG within TAG Contributor Strategy.

Let's review and discuss in our next Dev meeting?

PR #688, opened 2022-11-02
Add governance

dholbach:
Eventually fixes: #642


jackfrancis:
Did an initial read through, this looks pretty good, I'd say let's promote it from Draft and start getting more eyes on it

GOVERNANCE.md
```diff
@@ -0,0 +1,131 @@
...
+must be invited to such closed meetings, except for any Maintainer who is
+accused of a CoC violation.
+
+## CNCF Resources
+
+Any Maintainer may suggest a request for CNCF resources, either in the
+[mailing list](TODO: link to developer/maintainer mailing list), or during a
+meeting.  A simple majority of Maintainers approves the request.  The Maintainers
+may also choose to delegate working with the CNCF to non-Maintainer community
+members.
```
dholbach, line 104: Not sure if this section makes much sense for us?

GOVERNANCE.md
```diff
@@ -0,0 +1,131 @@
...
+While most business in Kured is conducted by "lazy consensus", periodically
+the Maintainers may need to vote on specific actions or changes.
+A vote can be taken on [the developer mailing list](TODO) or
+[the private Maintainer mailing list][mailing-list] for security or conduct
+matters. Votes may also be taken at [the developer meeting][meeting]. Any
+Maintainer may demand a vote be taken.
+
+Most votes require a simple majority of all Maintainers to succeed. Maintainers
+can be removed by a 2/3 majority vote of all Maintainers, and changes to this
+Governance require a 2/3 vote of all Maintainers.
```
dholbach, line 128: The Flux project chose to go with https://fluxcd.io/governance/#unanimity-decisions for "big decision" kind of things. What I like about it is:
- aiming for consensus by not allowing vetos
- setting a deadline so decisions are not blocked on people being on extended holidays, etc

dholbach: Process-wise we could issues in `/kubereboot/kured` for documenting vote-worthy things. WDYT?

jackfrancis: I like this approach. The difficulty comes if defining the boundary between "big decisions" and "decisions". Given that the document as a whole describes expected community behaviors that include trust and cooperation, I think we shouldn't have to worry about a strict definition, we can rely upon the "good vibes" part of this doc to be the foundation of the community, and the "strict rules" part secondary.

So +1 to your suggestions.

dholbach: If you check out https://fluxcd.io/governance/ a lot of "big decision" type stuff is spelled out. Almost all of it in the realms of governance. If we are talking about big code changes, we could adopt a lighter version of https://github.com/fluxcd/community/blob/main/PROCESS.md#proposal-process?

But all of this might as well just be added in the second iteration of governance?

GOVERNANCE.md
ckotzbauer: I'm also not sure about this section. Maybe a bit abstract.

ckotzbauer:
I think this is a very good document to start :+1:

GOVERNANCE.md
jackfrancis: Agree, let's ship with this removed.

GOVERNANCE.md
jackfrancis: +1 to formalizing this definition in a future version, if in practice the community will benefit from that. In the meantime it's worth leaning on common sense as an initial guidepost (IMO).

jackfrancis:
changes lgtm

ckotzbauer:
LGTM

ckotzbauer (approved):

jackfrancis (approved):
/lgtm

hiddeco (approved):

evrardjp (approved):
LGTM

PR #697, opened 2022-11-14
Fix location of Dockerfile for dependabot

dholbach:


dholbach:
> /lgtm

This command is not picked up by GitHub. Kured does not use prow.

ckotzbauer (approved):

Issue #702, opened 2022-11-15
New Annotation reboot-required


timo-42:
Hi

We have some workload in the cluster, which can not be moved at will and requires maintenance windows. It would be much easier for our tools to move stuff around if Kured would set an annotation like `weave.works/kured-reboot-required=true` and we must not fall back to access an Prometheus.

`weave.works/kured-most-recent-reboot-needed` does not work for us, since it is only set when the reboot process is already started. We need an indicator beforehand, so we don't schedule long running stuff on it and move single instances in their respective maintenance window from the node.

Is this something the community would be interested in?

jackfrancis:
not stale

PR #715, opened 2023-01-10
feat: Add `reboot-required` annotation

timo-42:
#702

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -30,13 +30,13 @@ import (
 	"github.com/google/shlex"
 
 	shoutrrr "github.com/containrrr/shoutrrr"
-	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/kubereboot/kured/pkg/alerts"
 	"github.com/kubereboot/kured/pkg/daemonsetlock"
 	"github.com/kubereboot/kured/pkg/delaytick"
 	"github.com/kubereboot/kured/pkg/taints"
 	"github.com/kubereboot/kured/pkg/timewindow"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
 var (
@@ -95,6 +95,8 @@ const (
 	KuredRebootInProgressAnnotation string = "weave.works/kured-reboot-in-progress"
 	// KuredMostRecentRebootNeededAnnotation is the canonical string value for the kured most-recent-reboot-needed annotation
 	KuredMostRecentRebootNeededAnnotation string = "weave.works/kured-most-recent-reboot-needed"
+	// KuredRebootRequiredAnnotation is the canonical string value for the kured reboot-required annotation
+	KuredRebootRequiredAnnotation string = "weave.works/reboot-required"
 	// EnvPrefix The environment variable prefix of all environment variables bound to our command line flags.
 	EnvPrefix = "KURED"
 )
@@ -520,12 +522,37 @@ func invokeReboot(nodeID string, rebootCommand []string) {
 	}
 }
 
-func maintainRebootRequiredMetric(nodeID string, sentinelCommand []string) {
+func maintainRebootRequiredMetricAndAnnotation(nodeID string, sentinelCommand []string) {
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	client, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		log.Fatal(err)
+	}
+
 	for {
 		if rebootRequired(sentinelCommand) {
 			rebootRequiredGauge.WithLabelValues(nodeID).Set(1)
+			if annotateNodes {
+				// Annotate this node to indicate that "I want to be rebooted!"
+				// so that other node maintenance tools can move stateful workloads(in their respective maintenance windows) away
+				annotations := map[string]string{KuredRebootRequiredAnnotation: "true"}
+				err := addNodeAnnotations(client, nodeID, annotations)
+				if err != nil {
+					continue
+				}
+			}
 		} else {
 			rebootRequiredGauge.WithLabelValues(nodeID).Set(0)
+			if annotateNodes {
+				err := deleteNodeAnnotation(client, nodeID, KuredRebootRequiredAnnotation)
+				if err != nil {
+					continue
+				}
+			}
 		}
 		time.Sleep(time.Minute)
 	}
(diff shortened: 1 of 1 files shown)


cmd/kured/main.go
```diff
@@ -652,11 +670,9 @@ func rebootAsRequired(nodeID string, rebootCommand []string, sentinelCommand []s
 			// And finally (3) if it has that annotation, to delete it.
 			// This indicates to other node tools running on the cluster that this node may be a candidate for maintenance
 			if annotateNodes && !rebootRequired(sentinelCommand) {
-				if _, ok := node.Annotations[KuredRebootInProgressAnnotation]; ok {
-					err := deleteNodeAnnotation(client, nodeID, KuredRebootInProgressAnnotation)
-					if err != nil {
-						continue
-					}
+				err := deleteNodeAnnotation(client, nodeID, KuredRebootInProgressAnnotation)
```
ckotzbauer, line 673: Why the if was removed? It would avoid a call to the apiserver.

ckotzbauer (changes requested):
Thanks for your PR. I added one smaller comment above.
I think it would make sense to only add/remove this annotation when the `--annotate-nodes` flag is set. This would streamline the new annotation with the others.

cmd/kured/main.go
timo-42: Readded the check.

timo-42:
> Thanks for your PR. I added one smaller comment above. I think it would make sense to only add/remove this annotation when the `--annotate-nodes` flag is set. This would streamline the new annotation with the others.

Put the new annotation behind the annotation toggle.

timo-42:
@ckotzbauer I pushed the changes you requested.

timo-42:
@ckotzbauer @dholbach Can you please have a look?

ckotzbauer:
@timo-42 I have not forgotten, but I am a bit busy at the moment.

timo-42:
You're right. I did not consider checking the state outside the window! Moved the code to a separate goroutine.
The question is: Should we create a separate goroutine for prometheus metric and the annotations or should we merge it?

vitality411:
I'm not a developer per se, but I think it would have advantages for the memory footprint to have only one goroutine. Especially since they partially execute the same commands.

timo-42:
Moved code into metric goroutine.

timo-42:
@vitality411 @ckotzbauer Is there anything more I should do?

vitality411:
Hi @timo-42, from my side it looks good.

jackfrancis:
@timo-42 Thank you for your patience during the review process, I'm back online and able to do this responsively going forward.

As I understand it, what you'd like to accomplish here is to not schedule certain workloads on nodes that are running kured, because those nodes will responsively reboot, and cordon/drain is not sufficient to gracefully evict and reschedule those workloads. Is that correct?

If so, I wonder if this issue would be a more elegant solution: https://github.com/kubereboot/kured/issues/737

My observation: there is a problem with using a node annotation via the kured daemonset to identify "I am running the kured daemonset", because if for any reason that daemonset is removed manually, and/or the appropriate node taints are added so that it is no longer running kured, that annotation will have to be manually removed (and inevitably this will not happen 100% of the time, and so that annotation will not be 100% accurate in reflecting the state of that node).

timo-42:
> As I understand it, what you'd like to accomplish here is to not schedule certain workloads on nodes that are running kured, because those nodes will responsively reboot, and cordon/drain is not sufficient to gracefully evict and reschedule those workloads. Is that correct?

Yes. We want to not schedule new workload on it and want to be notified that a node needs a reboot, so we can start draining the pods in their respective maintenance windows.

> If so, I wonder if this issue would be a more elegant solution: #737

If I understand this correctly, the proposal is to `run an arbitrary kubectl command`  before draining. This is already to late in the reboot process for us. We need days to drain a node, therefore we need the annotation as soon as possible and before the kured acquires the cluster lock and runs the reboot command.

> My observation: there is a problem with using a node annotation via the kured daemonset to identify "I am running the kured daemonset", because if for any reason that daemonset is removed manually, and/or the appropriate node taints are added so that it is no longer running kured, that annotation will have to be manually removed (and inevitably this will not happen 100% of the time, and so that annotation will not be 100% accurate in reflecting the state of that node).

If the concern is cleanup, than we could add a [pre delete hook](https://helm.sh/docs/topics/charts_hooks/#the-available-hooks) to delete all dangling annotations.

jackfrancis:
@timo-42 Thanks for the explanation.

Here's my thinking: by putting the annotation on the node as soon as kured launches its daemonset runtime you are essentially declaring: don't schedule any non-evictable workloads on this node because I *may* be rebooted at some point in the future. And then you would begin the process of evicting existing workloads, which may take days. But it's not clear to me how having this advance notice will actually guarantee that the underlying host OS will not put a `/var/run/reboot-required` file before that days-long eviction completes. And so you are still at risk of the node rebooting during the workload eviction process.

It would seem to me that the workload requirements you're outlining would forbid kured from running at all on the nodes that are running those workloads. What is the downside of simply tainting such nodes and not having kured run on it. I'm not sure that kured is as it currently is a viable solution for these operational requirements.

@dholbach @ckotzbauer @evrardjp any other thoughts?

timo-42:
@jackfrancis 
> It would seem to me that the workload requirements you're outlining would forbid kured from running at all on the nodes that are running those workloads. What is the downside of simply tainting such nodes and not having kured run on it. I'm not sure that kured is as it currently is a viable solution for these operational requirements.

This is easy :) - I use `--blocking-pod-selector` to prevent rebooting nodes which have these kinds of workloads.

jackfrancis:
@timo-42 Thanks for clarifying, that makes sense!

In this case this would be my suggestion (curious for others' thoughts as well):

Rather than add a third annotation to a node to describe kured reboot state, let's simply re-order the steps in the flow where we annotate `"weave.works/kured-reboot-in-progress"`. At present, we short-circuit the flow if a matching blocking pod selector is found, but we can short-circuit due to that critera _later_ in the flow. For example:

If a reboot is required:
1. Add the `"weave.works/kured-reboot-in-progress"` annotation
2. Add the "prefer no schedule" taint to the node
3. If reboot is blocked due to matching pod selector, continue
  a. Else, cordon/drain and reboot node

The above should solve for your use-case. You may use the `"weave.works/kured-reboot-in-progress"` annotation in your own workflows to prevent new, non-interruptible workloads from landing on this node (or you may be able to rely upon the "prefer no schedule" taint, but that taint is not strict, so FYI). Only when existing, non-interruptible workloads have gracefully exited, and the pod selector finds no match, will the node actually cordon/drain + reboot.

@ckotzbauer @dholbach @evrardjp if they have any criticisms to the above

ckotzbauer:
I like the idea @jackfrancis. I'm not fully aware of the current implementation of `"weave.works/kured-reboot-in-progress"`: Would this be a breaking change?

jackfrancis:
@ckotzbauer In my view it wouldn't be a breaking change. The original implementation of `"weave.works/kured-reboot-in-progress"` was meant to annotate the node *as soon as we know it will eventually be rebooted*. One particular condition that can delay a reboot from happening right away is the inability to acquire the exclusive node reboot lock. But we still want to annotate the node because in the next loop we may require that lock, and thus we still don't want applications that are consuming that annotation to consider that node as a viable node for scheduling new workloads.

It seems that the blocking pod selector criteria was overlooked when implementing the `"weave.works/kured-reboot-in-progress"` annotation, but that criteria is functionally equivalent to the exclusive node reboot lock: in the next loop there may be no more pods that match the blocking pod criteria, and so we are in a similar situation where we know that the node will eventually be rebooted.

tl;dr I think there's a bug in the implementation of the `"weave.works/kured-reboot-in-progress"` node annotation, and if we fix that bug, then @timo-42 and other folks can use _it_ to build conditional scheduling flows around.

@timo-42 does that make sense? I'd be happy to fix that bug myself.

timo-42:
Closing since #749 got merged.

PR #786, opened 2023-06-20
add argument to invert the behavior of alert-filter-regexp

jimliming:
Adding "--alert-filter-match-only" argument which inverts the behavior of "alert-filter-regexp". When alert-filter-match-only=true, kured will block on any active alerts that match the regexp defined in 'alert-filter-regexp'

A use case for this is: 
Fire alerts if the cluster is becoming unhealthy (a % of nodes are no longer == Ready). Immediately prevent kured from continuing to drain and reboot.

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -363,25 +363,10 @@ type KubernetesBlockingChecker struct {
 }
 
 func (pb PrometheusBlockingChecker) isBlocked() bool {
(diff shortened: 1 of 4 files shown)


ckotzbauer (dismissed):

ckotzbauer:
@jimliming I just pushed some small code-improvements and unit-tests to simplify things and to not ignore the `--alert-firing-only` flag.

ckotzbauer (approved):

jimliming:
makes sense - good catch on alert-firing-only

Issue #737, opened 2023-02-28
Custom drain command?


harrytang:
Hi there,

I have been using Kured and I find it to be a great tool for automating node reboots in a Kubernetes cluster.

I was wondering if there are any plans to add support for custom drain commands in Kured? It would be really helpful if we could specify our own custom drain command that Kured would execute before rebooting a node.

If this is not currently on the roadmap, I would love to know if it's something that the Kured development team would consider adding in the future.

Thank you for your time and for your work on Kured. I look forward to hearing back from you.

Best regards,
Harry

jackfrancis:
Hi Harry, we could be open to that. Could you provide an example of what you'd like to do in addition to (or instead of) the normal k8s "drain node" behavior?

harrytang:
Hi, 

We are currently using Longhorn Storage in our cluster, and while a node is being drained, we still need some components functioning so that the volumes can be properly detached. (see https://longhorn.io/docs/1.4.0/volumes-and-nodes/maintenance/#updating-the-node-os-or-container-runtime)

We normally use this drain command:
```bash
kubectl drain NODEX --delete-emptydir-data --ignore-daemonsets --pod-selector='app!=csi-attacher,app!=csi-provisioner,longhorn.io/component!=instance-manager'
```
Hope you find this helpful.

Thank1

jackfrancis:
That helps, thanks @harrytang!

I'll think about how we might put something like this together, stay tuned!

harrytang:
Github keep

ddsmith2-eprod:
Are there any additional suggestions on how to deal with this issue?   If you have volume that is not replicated the node fails to drain due to the pod disruption budget.  This happens over time when a volume is not used as much and you only have one replica.  I see that my predecessors used to stop dockerd and iscsid using Ansible to patch and reboot nodes.

I tried setting forceReboot=true, but it does not seem to help.

docbobo:
I am in the same boat as everyone else regarding Longhorn draining. "Allow Node Drain with the Last Healthy Replica" is not solving this issue for me though.

tylerauerbeck:
Is anybody taking a look at this? I'm currently attempting to find ways of creating alert manager silences and then removing them when the node comes back up, so I'd want to have a pre reboot command and a post reboot command just like there are for labels. I'd imagine it could tie into the same hooks that the labels use and take a similar approach to how users are able to specify their own reboot command. 

[...]

PR #788, opened 2023-06-27
Support pod-selector for drain command

docbobo:
This PR is related to issue #737. While it does not add support for custom drain commands, it adds support for configuring the pod-selector to be used for draining, which should be sufficient to address all the Longhorn related issues.

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -49,6 +49,7 @@ var (
 	period                          time.Duration
 	metricsPort                     int
 	drainGracePeriod                int
+	drainPodSelector                string
 	skipWaitForDeleteTimeoutSeconds int
 	dsNamespace                     string
 	dsName                          string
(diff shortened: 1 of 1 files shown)


cmd/kured/main.go
```diff
@@ -129,6 +130,8 @@ func NewRootCommand() *cobra.Command {
 		"port number where metrics will listen")
 	rootCmd.PersistentFlags().IntVar(&drainGracePeriod, "drain-grace-period", -1,
 		"time in seconds given to each pod to terminate gracefully, if negative, the default value specified in the pod will be used")
+	rootCmd.PersistentFlags().StringVar(&drainPodSelector, "drain-pod-selector", "",
+		"label selector to filter pods on the node when draining (default: '')")
```
jackfrancis, line 134: Can we make it clear that this is an "inclusive" filter? Basically by using a pod-selector you are instructing drain to "delete only these pods that match this label selector".

Otherwise "filter" can suggest "filter out these pods", i.e., *don't* delete these pods.

jackfrancis:
@docbobo Also make sure to sign your commit (see https://github.com/kubereboot/kured/pull/788/checks?check_run_id=14582462611 for more info)

docbobo:
Okay, tried to address both issues. Please have another look at the description and let me know if you want additional changes.

jackfrancis:
code lgtm, looks like it's still complaining about an unsigned commit

docbobo:
Yes, I noticed. But there is only one commit in this PR and that's verified. Maybe the result of the required rebasing? Do I need to reopen the PR?

docbobo:
Nevermind. The commit was signed but missing the Signed-off-by. Now it should be correct.

jackfrancis (approved):
lgtm

@docbobo could you submit a helm chart change to expose this new feature?

https://github.com/kubereboot/charts

We'll merge that once this new feature is part of a released version of kured.

ckotzbauer (approved):

PR #808, opened 2023-08-02
Use Background context

### LINKED ISSUE #234: Use of context.TODO()
I noticed that you added context arguments to the k8s API calls in 16109017cec02a8a0c4f80f829acc38ab28a14b2 - but you used context.TODO() instead of propagating a context object.

[context.TODO()](https://golang.org/pkg/context/#TODO) signals that this will be changed at some point - do have any plans to do so?
Otherwise, it would be more appropriate to use [context.Background()](https://golang.org/pkg/context/#Background) instead.

ckotzbauer:
close #234


jackfrancis (changes requested):
I don't think we wanna bother with this. My thoughts here:

- https://github.com/kubereboot/kured/issues/234#issuecomment-1681319739

PR #811, opened 2023-08-03
Adds new flag --metrics-host

nkinkade:
Kured runs with `hostNetwork=true`, which means that, today, Kured will expose the metrics endpoint on what is probably a public interface for many deployments. The metrics are generally not sensitive data, but as a policy it seems wrong to do this. Allowing the user to configure which port Kured exposes metrics on is a step in the right direction, but the current implementation does not allow the user to specify the address to listen on. For example, in our use case, we would like for the metrics server to _only_ listen on a loopback address. This would allow us to put [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) in front of the metrics endpoint.

This change could slightly complicate the Helm chart because it not only affects the metrics endpoint, but also the `readinessProbe` configuration.

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -47,7 +47,7 @@ var (
 	drainTimeout                    time.Duration
 	rebootDelay                     time.Duration
 	period                          time.Duration
-	metricsAddress                  string
+	metricsPort                     int
 	drainGracePeriod                int
 	drainPodSelector                string
 	skipWaitForDeleteTimeoutSeconds int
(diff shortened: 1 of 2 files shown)


nkinkade:
I just noticed that a PR was merged yesterday which set `hostNetwork=false` by default in the chart. That change may obviate the need for this PR, at least for our use case. However, in cases where someone does for some reason want to run with `hostNetwork=true` it may still be useful for them to be able to specify the full listen address. There may be other reasons I'm not thinking of.

ckotzbauer (changes requested):
Hi @nkinkade,
thanks for your PR. Yes, since a few days, the `hostNetwork=true` setting is not the default anymore, so this problem should not occur too often.

But, you are right, it may be good in some cases to also configure the host of the interface to listen on. However, the `--metrics-port` flag was released with `1.13.2`, so this PR would be a breaking change, which we want to avoid. Instead of replacing the flag it would be good to add an additional `--metrics-host` flag to kured (and also to the chart).

nkinkade:
@ckotzbauer: I have modified the PR and its title. The changes now only add a new flag `--metrics-host`.

I am happy to contribute a corresponding PR to the charts repo, once I have confirmation from you that this change looks okay.

Thank you.

ckotzbauer:
The code looks good, thanks!
Can you please
- signoff all commits to our repos?
- open a PR against the chart-repository
- open a PR against the website-repository and document the flag

Thanks!!

nkinkade:
@ckotzbauer:

* I believe I have now signed all commits in this PR
* https://github.com/kubereboot/charts/pull/50
* https://github.com/kubereboot/website/pull/69

Let me know if I missed anything.

Thank you!

ckotzbauer (approved):

ckotzbauer:
Thanks for the changes!!

PR #813, opened 2023-08-04
Sentinel-command without nsenter by default

### LINKED ISSUE #526: improve security of sentinel mechanism
The default sentinel behaviour is to check for the existence of a file on the host. This is done using a `test -f` command executed on the host with nsenter:
https://github.com/weaveworks/kured/blob/main/cmd/kured/main.go#L661

I would propose refactoring this to achieve the same thing in a different way, using a (read-only) hostPath to check for existence of a file on the host, while keeping the non-default option to run an arbitrary sentinel command in the host namespace unchanged. In conjunction with #416  this would allow kured to be non-privileged.  This could mostly be handled by the Helm chart, mounting a configurable directory on the host (the sentinel file itself could not be mounted as a file-type hostPath because it would normally not exist).

```
volumes: 
  - name: sentinel
    hostPath:
      path: /somewhere_configurable
      type: Directory
```      
      
```  
volumeMounts:
  - name: sentinel
    mountPath: /somewhere_arbitrary
    readOnly: true
```

A rebootSentinel option already exists in the helm chart and kured CLI.
Looks like these helm [path helpers](https://helm.sh/docs/chart_template_guide/accessing_files/#path-helpers) would be perfect to split up the given rebootSentinel path; `Dir` could be used to get the hostPath directory to mount and `Base` could be passed to the kured executable, to find the file name inside the arbitrary directory in the pod.

ckotzbauer:
As long as there's no custom sentinel-command kured expects that the sentinel file (`/var/run/reboot-required` by default) to be mounted into the container. This would cause the sentinel-command to be executed without nsenter.

close #526

### DIFF
--- a/kured-ds.yaml
+++ b/kured-ds.yaml
@@ -56,9 +56,7 @@ spec:
           volumeMounts:
             - mountPath: /sentinel
               name: sentinel
-              # This can't be read-only because the service-account-token is mounted within the same directory
-              # This would cause a crash at startup
-              readOnly: false
+              readOnly: true
           command:
             - /usr/bin/kured
             - --reboot-sentinel=/sentinel/reboot-required
--- a/kured-ds.yaml
+++ b/kured-ds.yaml
@@ -56,7 +56,9 @@ spec:
           volumeMounts:
             - mountPath: /sentinel
               name: sentinel
-              readOnly: true
+              # This can't be read-only because the service-account-token is mounted within the same directory
+              # This would cause a crash at startup
+              readOnly: false
           command:
             - /usr/bin/kured
             - --reboot-sentinel=/sentinel/reboot-required


rptaylor:
@ckotzbauer  Great!  The kube-api-access projected volume is also read-only BTW, but indeed overlapping volumeMounts would not be good.

Fortunately I think there is a relatively easy way to allow the hostPath to be read-only as a security improvement.
The sentinel path is a single configurable value currently interpreted as one path but we need a way to use it for two different purposes: the path on the host (needs to be configurable) and the path in the container (can be arbitrary), since hostPath can map between different paths.

Moreover I think this can be done entirely in Helm, without further go changes. I can put together a PR to the Helm chart illustrating the idea. It would not require any changes to kured-ds.yaml; is that just an example or for people who want to manually apply static YAML?

ckotzbauer:
Yeah, you are absolutely right, I thought about the same approach today. We could leave the default sentinel-path in go as is and just map in the helm-chart to an arbitrary path which does not overlap. This would be absolutely feasible, but it would make kustomize-based installations a bit more complicated, as you have to configure the deployment properly.
That's what you mentioned, the `kured-ds.yaml` is not just an example it is the "trusted source" for all which are not using helm to install kured.

But however, maybe this would be a good improvement and we should just take the risk of making the kustomize-setups a bit more complicated. We would just need to adjust my PR kubereboot/charts#49

rptaylor:
Actually I noticed the chart already allows arbitrary volume(Mounts) to be defined, so it could be done without any further modifications to the chart at all.  This is what I would do in my Helm values:

```
volumes:
  - hostPath:
    name: sentinel
      path: /opt/whatever/
      type: Directory
      read-only: true
volumeMounts:
  - name: sentinel
    mountPath: /kured

configuration:
  rebootSentinel: /kured/reboot-required
  rebootSentinelCommand: ""
```

Then I place a sentinel file in /opt/whatever/reboot-required, and kured will look for the hostPath-mapped location /kured/reboot-required.
As long as the empty sentinel command causes kured to look for the file via 'test -f' without nsenter, it will work.

If the Helm chart is expanded via something like https://github.com/kubereboot/charts/pull/49, I think it would help make the configuration easier, but it should be done in a way that makes the hostPath read-only by default, since the goal is to improve security. That would require using some Helm path helpers as I mentioned here which would be a bit more complicated: https://github.com/kubereboot/kured/issues/526#issue-1192414701   But I would be happy with just defining the volumes via my values. We could write a simple doc or add some comments explaining how to do this.

ckotzbauer:
The `kured-ds.yaml` and the helm-chart are updated with a `/sentinel` read-only mount.

kured-ds.yaml
```diff
@@ -48,8 +53,13 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
+          volumeMounts:
+            - mountPath: /sentinel
```
jackfrancis, line 57: Are we using `/sentinel` as the path inside the container so that we still have access to the container-specific `/var/run` directory? If we don't care about that, using a different name like `sentinel` as the mount path inside the container could confuse some folks.

I see that you and @rptaylor have had a conversation around this, so if you feel O.K. about changing this going forward, I'm lgtm as-is.

The obvious gotcha is that existing, static daemonset files that people are using won't work after we make this change (`/var/run/reboot-required` will point to the local container and not the underlying host filesystem), so they will need to update any of those files to accepted the changes we've added to our `kured-ds.yaml` reference here.

jackfrancis (approved):
lgtm pending consensus on concerns about renaming the container mount path

kured-ds.yaml
rptaylor: The path inside the container is arbitrary, since the hostpath volume can be mounted anywhere. The helm template takes care of mounting volumes in the right place and giving the right path arg to kured in a way that is transparent to the user. The user only needs to specify the location of the sentinel on the host.

I'm not sure I follow. The kured-ds.yml file uses a hostPath to mount /var/run/reboot-required to /sentinel/reboot-required and tells kured to use that location, so there should be no change in behaviour for users.

ckotzbauer: I see your point @jackfrancis. Not all users which use the static manifest may update the whole manifest all time, but instead only update the tag. We will break those users, but I think it should be enough to mention this clearly in the changelog.

rptaylor: It looks like this MR has been approved and could be merged? Or is there more to be discussed/agreed here first? @jackfrancis @ckotzbauer 
Thanks!

ckotzbauer:
@rptaylor Yes, this is ready to merge. I will hold it back until we know when we do the next minor release (around Kubernetes 1.29.0). Until that we can do patch releases from main without separate branching.

rptaylor:
This will be superseded by https://github.com/kubereboot/kured/pull/814  right?

ckotzbauer:
@rptaylor No, they are both needed. This PR is "only" needed to make the check if a reboot is required more secure.

rptaylor:
Okay I wasn't sure because the commit descriptions in this PR are a subset of those in https://github.com/kubereboot/kured/pull/814/commits.  Maybe once that is merged and this is rebased it will be more clear. Thanks!

jackfrancis (dismissed):
lgtm

ckotzbauer:
Superseded by #814

PR #814, opened 2023-08-05
Add signal-reboot

### LINKED ISSUE #416: improve security of reboot mechanism
Hello,

We would really like to use kured but there is reluctance about a fully privileged daemonset that has access to execute arbitrary commands as root on every node.  I have reviewed some options that could allow rebooting more securely.

- **Option 0** (current kured behaviour):  nsenter and execute a configurable shutdown command (`/bin/systemctl reboot` by default) in the host namespace to reboot. Requires full privileges.
- **Option 1** grant CAP_SYS_BOOT to kured to allow the [reboot(2)]( https://man7.org/linux/man-pages/man2/reboot.2.html) syscall
  - **Option 1A** semi-dirty reboot:  call sync() and then reboot(). Should avoid any filesystem corruption, but systemd will not be aware of the reboot so it is not very friendly to system services. Was discussed in https://github.com/weaveworks/kured/pull/172
  - **Option 1B** graceful reboot via Ctrl Alt Del: invoke reboot syscall with LINUX_REBOOT_CMD_CAD_OFF, then issue a CtrlAltDel keystroke to trigger a graceful systemd-managed shutdown via the ctrl-alt-del.target unit file.
- **Option 2**: graceful reboot via a kill signal. Grant CAP_KILL to kured so it can send SIGRTMIN+5 to PID 1 (systemd), which is [an equivalent way](https://man7.org/linux/man-pages/man1/systemd.1.html#SIGNALS) to achieve a graceful systemd-managed reboot (without relying on the ctrl-alt-del.target unit, thanks to evrardjp's comment - though this may be less portable than SIGINT) .

Comparisons
- Option 0
  -  pros: currently working this way
  -  cons: not very secure, pod runs with far more privileges than should be needed to reboot
- Option 1A:
  - pros: easy to do, most secure (only CAP_SYS_BOOT required)
  - cons: not graceful (systemd-managed)
- Option 1B:
  - would be the best of both worlds but after further investigation I am not sure if/how it is possible. Here is [example code](https://github.com/micmonay/keybd_event/blob/master/keybd_linux.go) to issue a CtrlAltDel keystroke but using uinput which requires root privilege so that does not help. It might be possible with ioctl but that could require additional capabilities (using TIOCSTI, CAP_SYS_ADMIN?). (Also users may have masked ctrl-alt-del.target or changed the reboot keystroke from the default CtrlAltDel)
- Option 2:
  - pros: easy to do, graceful, requires far fewer privileges (only CAP_KILL), more secure than Option 0
  - cons: sending any signal to any host process is still a significant attack surface. 
  
 All the options still require hostPID: true. 
  
Anyway do you think option 2 is at least a clear improvement over 0 and might be suitable as a new default behaviour? And maybe Option 1 A (or B if possible) could be configurable as more secure alternatives?

### LINKED ISSUE #722: Kured pods need to run without privileged permissions
image
 
Facing this issue when i try to run this daemonset as a non-privilleged permission.

Please anyone from team help me on this

ckotzbauer:
Based on #813
close #416 
close #722

This PR adds a `--reboot-method` flag with "command" (default) and "signal" option. The "command" option uses the `--reboot-command` on the host with `nsenter` as before. The new "signal" mode uses a `SIGRTMIN+5` signal by default against PID 1 (systemd) to reboot the node. The signal can be changed via `--reboot-signal` flag.

With this, the kured pod runs without privileged mode.

This PR is published as docker-image (amd64 and arm64): `ghcr.io/ckotzbauer/kured:1.14.0-alpha.2`
Usage (based on the latest helm-chart) - Helm values.yaml:

```yaml
image:
  repository: ghcr.io/ckotzbauer/kured
  tag: 1.14.0-alpha.2
updateStrategy: RollingUpdate
configuration:
  period: "0h0m30s"
  rebootDelay: 0h1m0s
  rebootSentinel: /sentinel/reboot-required
extraArgs:
  reboot-method: signal
containerSecurityContext:
  readOnlyRootFilesystem: true
  privileged: false
  allowPrivilegeEscalation: false
  capabilities:
    drop: ["*"]
    add: ["CAP_KILL"]
volumes:
  - name: sentinel
    hostPath:
      path: /var/run
      type: Directory
volumeMounts:
  - name: sentinel
    mountPath: /sentinel
    readOnly: true
```

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -77,6 +77,7 @@ var (
 	messageTemplateUncordon         string
 	podSelectors                    []string
 	rebootCommand                   string
+	rebootSignal                    int
 	logFormat                       string
 	preRebootNodeLabels             []string
 	postRebootNodeLabels            []string
@@ -186,6 +187,8 @@ func NewRootCommand() *cobra.Command {
 		"command to run when a reboot is required")
 	rootCmd.PersistentFlags().IntVar(&concurrency, "concurrency", 1,
 		"amount of nodes to concurrently reboot. Defaults to 1")
+	rootCmd.PersistentFlags().IntVar(&rebootSignal, "reboot-signal", 34+5,
+		"signal to use for reboot, SIGRTMIN+5 by default.")
 
 	rootCmd.PersistentFlags().StringVar(&slackHookURL, "slack-hook-url", "",
 		"slack hook URL for reboot notifications [deprecated in favor of --notify-url]")
@@ -562,7 +565,7 @@ func invokeReboot(nodeID string, rebootCommand []string) {
 	if rebootMethod == MethodCommand {
 		booter = reboot.NewCommandReboot(nodeID, rebootCommand)
 	} else if rebootMethod == MethodSignal {
-		booter = reboot.NewSignalReboot(nodeID)
+		booter = reboot.NewSignalReboot(nodeID, rebootSignal)
 	} else {
 		log.Fatalf("Invalid reboot-method configured: %s", rebootMethod)
 	}
@@ -872,8 +875,10 @@ func root(cmd *cobra.Command, args []string) {
 	log.Infof("Reboot check command: %s every %v", sentinelCommand, period)
 	log.Infof("Concurrency: %v", concurrency)
 	log.Infof("Reboot method: %s", rebootMethod)
-	if rebootCommand == MethodSignal {
+	if rebootCommand == MethodCommand {
 		log.Infof("Reboot command: %s", restartCommand)
+	} else {
+		log.Infof("Reboot signal: %v", rebootSignal)
 	}
 
 	if annotateNodes {
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -880,8 +880,13 @@ func root(cmd *cobra.Command, args []string) {
 	// To run those commands as it was the host, we'll use nsenter
 	// Relies on hostPID:true and privileged:true to enter host mount space
 	// PID set to 1, until we have a better discovery mechanism.
-	hostSentinelCommand := buildHostCommand(1, sentinelCommand)
 	hostRestartCommand := buildHostCommand(1, restartCommand)
+
+	// Only wrap sentinel-command with nsenter, if a custom-command was configured, otherwise use the host-path mount
+	hostSentinelCommand := sentinelCommand
+	if rebootSentinelCommand != "" {
+		hostSentinelCommand = buildHostCommand(1, sentinelCommand)
+	}
 
 	go rebootAsRequired(nodeID, hostRestartCommand, hostSentinelCommand, window, lockTTL, lockReleaseDelay)
 	go maintainRebootRequiredMetric(nodeID, hostSentinelCommand)
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -33,8 +33,10 @@ import (
 	"github.com/kubereboot/kured/pkg/alerts"
 	"github.com/kubereboot/kured/pkg/daemonsetlock"
 	"github.com/kubereboot/kured/pkg/delaytick"
+	"github.com/kubereboot/kured/pkg/reboot"
 	"github.com/kubereboot/kured/pkg/taints"
 	"github.com/kubereboot/kured/pkg/timewindow"
+	"github.com/kubereboot/kured/pkg/util"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
@@ -47,6 +49,7 @@ var (
 	drainDelay                      time.Duration
 	drainTimeout                    time.Duration
 	rebootDelay                     time.Duration
+	rebootMethod                    string
 	period                          time.Duration
 	metricsHost                     string
 	metricsPort                     int
@@ -103,6 +106,11 @@ const (
 	KuredMostRecentRebootNeededAnnotation string = "weave.works/kured-most-recent-reboot-needed"
 	// EnvPrefix The environment variable prefix of all environment variables bound to our command line flags.
 	EnvPrefix = "KURED"
+
+	// MethodCommand is used as "--reboot-method" value when rebooting with the configured "--reboot-command"
+	MethodCommand = "command"
+	// MethodSignal is used as "--reboot-method" value when rebooting with a SIGRTMIN+5 signal.
+	MethodSignal = "signal"
 )
 
 func init() {
@@ -146,6 +154,8 @@ func NewRootCommand() *cobra.Command {
 		"timeout after which the drain is aborted (default: 0, infinite time)")
 	rootCmd.PersistentFlags().DurationVar(&rebootDelay, "reboot-delay", 0,
 		"delay reboot for this duration (default: 0, disabled)")
+	rootCmd.PersistentFlags().StringVar(&rebootMethod, "reboot-method", "command",
+		"method to use for reboots. Available: command")
 	rootCmd.PersistentFlags().DurationVar(&period, "period", time.Minute*60,
 		"sentinel check period")
 	rootCmd.PersistentFlags().StringVar(&dsNamespace, "ds-namespace", "kube-system",
@@ -299,22 +309,6 @@ func flagToEnvVar(flag string) string {
 	return fmt.Sprintf("%s_%s", EnvPrefix, envVarSuffix)
 }
 
-// newCommand creates a new Command with stdout/stderr wired to our standard logger
-func newCommand(name string, arg ...string) *exec.Cmd {
-	cmd := exec.Command(name, arg...)
-	cmd.Stdout = log.NewEntry(log.StandardLogger()).
-		WithField("cmd", cmd.Args[0]).
-		WithField("std", "out").
-		WriterLevel(log.InfoLevel)
-
-	cmd.Stderr = log.NewEntry(log.StandardLogger()).
-		WithField("cmd", cmd.Args[0]).
-		WithField("std", "err").
-		WriterLevel(log.WarnLevel)
-
-	return cmd
-}
-
 // buildHostCommand writes a new command to run in the host namespace
 // Rancher based need different pid
 func buildHostCommand(pid int, command []string) []string {
@@ -327,7 +321,7 @@ func buildHostCommand(pid int, command []string) []string {
 }
 
 func rebootRequired(sentinelCommand []string) bool {
-	cmd := newCommand(sentinelCommand[0], sentinelCommand[1:]...)
+	cmd := util.NewCommand(sentinelCommand[0], sentinelCommand[1:]...)
 	if err := cmd.Run(); err != nil {
 		switch err := err.(type) {
 		case *exec.ExitError:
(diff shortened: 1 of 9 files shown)


rptaylor:
In my test this works, using the same values.yml file:
```
$ helm -n kured install -f ~/kured/values.yml   kured kubereboot/kured
$ helm -n kured list
NAME 	NAMESPACE	REVISION	UPDATED                                	STATUS  	CHART      	APP VERSION
kured	kured    	1       	2023-08-08 21:50:33.438477295 +0000 UTC	deployed	kured-5.1.0	1.13.2     
$ kubectl -n kured exec -it kured-vs4xb -- /bin/sh
/ # kill -s SIGRTMIN+5   1
```
Then journald shows:
```
Aug 08 14:52:08 kermes-dev-k8s-node-a03 systemd[1]: Received SIGRTMIN+6 from PID 22396 (sh).
```
and the system immediately does a clean reboot. However note that "SIGRTMIN+5" was sent but it was interpreted by host as SIGRTMIN+6.  I believe that is related to the different output of `kill -l` in the pod vs on the host, which is a potential compatibility issue that we want to avoid by not using the pod's kill executable in the first place; the go library will send the signal instead.  When I instead did `kill -s 39 1`1 it worked as expected.  I'm not sure how universal the interpretation 39=SIGRTMIN+5 is or whether it could be a good idea for the precise signal to be configurable.

Anyway this was running on a v1.23 k8s cluster with the old PSP admission controller disabled. I'm afraid I could not reproduce the issue; it works fine for me.  Looking at the capabilities of the kured process it also appears fine:
```
/ # ps -ef | grep kured | grep -v grep
 6107 root      0:00 /usr/bin/kured --ds-name=kured --ds-namespace=kured --metrics-port=8080 --period=0h0m30s --reboot-sentinel=/sentinel/reboot-required --reboot-command=/bin/systemctl reboot --reboot-delay=0h1m0s --log-format=text --reboot-method=signal
/ # grep ^Cap /proc/6107/status
CapInh:	0000000000000000
CapPrm:	00000000a80425fb
CapEff:	00000000a80425fb
CapBnd:	00000000a80425fb
CapAmb:	0000000000000000
```

Also when I touched the sentinel file on the host, kured worked as expected and the node rebooted:
```
time="2023-08-08T22:27:19Z" level=info msg="Reboot required"
.....   #evicting pods etc
time="2023-08-08T22:27:58Z" level=info msg="Delaying reboot for 1m0s"
time="2023-08-08T22:28:58Z" level=info msg="Emit reboot-signal for node: cluster-dev-k8s-node-a03"
time="2023-08-08T22:28:58Z" level=info msg="Waiting for reboot"
```

ckotzbauer:
Thanks for your tests and the positive feedback!

Yes, the specific signal value may indeed be a problem. We would need to add a good guidance for the user in our docs, to determine which signal value is correct for their host OS. Most of the users won't know as this is expert knowlege :laughing:

ckotzbauer:
I made it work on my cluster with `priviliged: false`, by disabling AppArmor for the container with the Pod-Annotation
```
container.apparmor.security.beta.kubernetes.io/kured
```

rptaylor:
Ah of course, I'm not familiar with AppArmor at all but that is the SELinux equivalent on some OSes.

cmd/kured/main.go
```diff
@@ -173,6 +184,8 @@ func NewRootCommand() *cobra.Command {
 		"command to run when a reboot is required")
 	rootCmd.PersistentFlags().IntVar(&concurrency, "concurrency", 1,
 		"amount of nodes to concurrently reboot. Defaults to 1")
+	rootCmd.PersistentFlags().IntVar(&rebootSignal, "reboot-signal", 34+5,
```
jackfrancis, line 187: let's convert the int sum literal into a const, like:

```
const (
...
    sigTrminPlus5 = 34 + 5`
...
)
```

cmd/kured/main.go
```diff
@@ -550,17 +547,22 @@ func uncordon(client *kubernetes.Clientset, node *v1.Node) error {
...
-
 	if notifyURL != "" {
 		if err := shoutrrr.Send(notifyURL, fmt.Sprintf(messageTemplateReboot, nodeID)); err != nil {
 			log.Warnf("Error notifying: %v", err)
 		}
 	}
 
-	if err := newCommand(rebootCommand[0], rebootCommand[1:]...).Run(); err != nil {
-		log.Fatalf("Error invoking reboot command: %v", err)
+	var booter reboot.Reboot
```
jackfrancis, line 556: I think the cleanest way to evolve the current implementation with this new code is to instantiate our `reboot.Rooter` object in `func root(cmd *cobra.Command, args []string)` and then pass that into `go rebootAsRequired(...)`

That would require the following modifications:

`rebootAsRequired` becomes:

`func rebootAsRequired(nodeID string, booter reboot.Reboot, sentinelCommand []string, window *timewindow.TimeWindow, TTL time.Duration, releaseDelay time.Duration) {`

We invoke `rebootAsRequired` thusly:

`go rebootAsRequired(nodeID, booter, hostSentinelCommand, window, lockTTL, lockReleaseDelay)`

In the flow of `rebootAsRequired`, instead of:

```
                invokeReboot(nodeID, rebootCommand)
		for {
			log.Infof("Waiting for reboot")
			time.Sleep(time.Minute)
		}
```

We do this:

```
                booter.Reboot()
		for {
			log.Infof("Waiting for reboot")
			time.Sleep(time.Minute)
		}
```

And now we can delete the `invokeReboot` func entirely.

What do you think?

jackfrancis:
@rptaylor thanks for your patience, added some review comments for @ckotzbauer, I think this generally looks like a great evolution, thanks this work!

ckotzbauer:
Sorry for the very long delay, this is ready for another review @jackfrancis @rptaylor

rptaylor:
@ckotzbauer very exciting! I can test it on a dev cluster, is there a new image available in ghcr.io/ckotzbauer/  ?

ckotzbauer:
Thanks for the hint: `ghcr.io/ckotzbauer/kured:1.15.0-alpha.0` (amd64 and arm64) @rptaylor

rptaylor:
Okay refreshing my memory on what this is even about 4 months later .... ;p

On a dev cluster I temporarily disabled PSP admission control for testing purposes.
```
$ cat values.yml 
image:
  repository: ghcr.io/ckotzbauer/kured
  tag: 1.15.0-alpha.0
configuration:
  period: "0h0m30s"
  rebootDelay: 0h1m0s
  rebootSentinel: /sentinel/reboot-required
# update faster
maxUnavailable: 10
extraArgs:
  reboot-method: signal
containerSecurityContext:
  readOnlyRootFilesystem: true
  privileged: false
  allowPrivilegeEscalation: false
  capabilities:
    drop: ["*"]
    add: ["CAP_KILL"]
volumes:
  - name: sentinel
    hostPath:
      path: /var/run
      type: Directory
volumeMounts:
  - name: sentinel
    mountPath: /sentinel
    readOnly: true
```

```
$ helm -n kured install -f values.yml   kured kubereboot/kured
$ helm -n kured list
NAME 	NAMESPACE	REVISION	UPDATED                                	STATUS  	CHART      	APP VERSION
kured	kured    	1       	2023-12-15 22:20:49.185347412 +0000 UTC	deployed	kured-5.2.0	1.14.0     
```
Then on a node I touched /var/run/reboot-required. 

pod logs:
```
$ kubectl -n kured logs -f kured-79vrp
time="2023-12-15T22:37:53Z" level=info msg="Binding node-id command flag to environment variable: KURED_NODE_ID"
time="2023-12-15T22:37:53Z" level=info msg="Kubernetes Reboot Daemon: 544a287"
time="2023-12-15T22:37:53Z" level=info msg="Node ID: cluster-dev-k8s-node-a02"
time="2023-12-15T22:37:53Z" level=info msg="Lock Annotation: kured/kured:weave.works/kured-node-lock"
time="2023-12-15T22:37:53Z" level=info msg="Lock TTL not set, lock will remain until being released"
time="2023-12-15T22:37:53Z" level=info msg="Lock release delay not set, lock will be released immediately after rebooting"
time="2023-12-15T22:37:53Z" level=info msg="PreferNoSchedule taint: "
time="2023-12-15T22:37:53Z" level=info msg="Blocking Pod Selectors: []"
time="2023-12-15T22:37:53Z" level=info msg="Reboot schedule: SunMonTueWedThuFriSat between 00:00 and 23:59 UTC"
time="2023-12-15T22:37:53Z" level=info msg="Reboot check command: [test -f /sentinel/reboot-required] every 30s"
time="2023-12-15T22:37:53Z" level=info msg="Concurrency: 1"
time="2023-12-15T22:37:53Z" level=info msg="Reboot method: signal"
time="2023-12-15T22:37:53Z" level=info msg="Reboot signal: 39"
time="2023-12-15T22:39:41Z" level=info msg="Reboot required"
time="2023-12-15T22:39:41Z" level=info msg="Acquired reboot lock"
time="2023-12-15T22:39:41Z" level=info msg="Draining node cluster-dev-k8s-node-a02"
evicting pod  ....
time="2023-12-15T22:40:35Z" level=info msg="Delaying reboot for 1m0s"
time="2023-12-15T22:41:35Z" level=info msg="Emit reboot-signal for node: cluster-dev-k8s-node-a02"
time="2023-12-15T22:41:35Z" level=info msg="Waiting for reboot"
```

Eventually ....  node rebooted:
```
Dec 15 14:41:35 cluster-dev-k8s-node-a02 systemd[1]: Received SIGRTMIN+5 from PID 2802298 (kured).
```

rptaylor:
Actually wait ... the node powered off instead of rebooting, that's not good...

rptaylor:
I tried again and the node rebooted normally. Sometimes our cloud infrastructure is a bit flakey... Anyway as long as systemd on the node reports "Received SIGRTMIN+5" it should be good.

cmd/kured/main.go
```diff
@@ -103,6 +107,13 @@ const (
 	KuredMostRecentRebootNeededAnnotation string = "weave.works/kured-most-recent-reboot-needed"
 	// EnvPrefix The environment variable prefix of all environment variables bound to our command line flags.
 	EnvPrefix = "KURED"
+
+	// MethodCommand is used as "--reboot-method" value when rebooting with the configured "--reboot-command"
+	MethodCommand = "command"
+	// MethodSignal is used as "--reboot-method" value when rebooting with a SIGRTMIN+5 signal.
+	MethodSignal = "signal"
+
+	sigTrminPlus5 = 34 + 5
```
rptaylor, line 116: Is this meant to be sigRtminPlus5  ?

ckotzbauer: Yes, I fix the typo.

ckotzbauer:
Thanks for testing @rptaylor, this should cover all needed changes, so this should be ready for merge.

jackfrancis (approved):
lgtm

great work, thank you!

cmd/kured/main.go
rptaylor: Looks like the typo is still there? But otherwise this should be finally good to go!

jackfrancis: Sorry, I didn't catch that. Let's fix the typo. If we're going to use camel case I'd vote for:

- `sigRTMin`

Maybe:

```go
const (
...
    sigRTMin = 34
    defaultSignal = sigRTMin + 5
...
)
```

And we should probably change the `MethodSignal` comment so that it doesn't imply that SIGRTMIN+5 is static (it's just a default, but the user can use whatever number within the min-max range that their OS allows). Maybe:

```
// MethodSignal is used as "--reboot-method" value when rebooting according to a signal in the allowed range between SIGRTMIN and SIGRTMAX as defined on the OS platform running kured
```

PR #835, opened 2023-09-20
Sample implementation of using Kubernetes leases instead of daemonset lock

### LINKED ISSUE #25: lock on a separate/safe object rather than DaemonSet
Current locking approach requires the kured job to have the RBAC privilege to be able to update its own DaemonSet.  This means a compromised kured job could redefine itself to (eg) elevate host access further(*).

Better (from an RBAC pov) would be to use a "harmless" resource type to store the lock annotation.  A ConfigMap (empty/dedicated for this purpose) would be ideal.

(*) I acknowledge that kured also needs other scary privileges (patch nodes) in order to conduct a drain/uncordon, and these won't be improved by this suggestion.  Small steps.

jmcshane:
Resolves #25

This is a demo implementation that I have working in my local environment. We can't directly use the `leaderelection` package from client-go without some modification (the daemonset will persist post restart of the node, so the original context of leader election will be lost).

Would love to chat about this implementation and get it ready for inclusion in this library. Thanks for your work on this great tool!

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -31,8 +31,8 @@ import (
 
 	shoutrrr "github.com/containrrr/shoutrrr"
 	"github.com/kubereboot/kured/pkg/alerts"
-	"github.com/kubereboot/kured/pkg/daemonsetlock"
 	"github.com/kubereboot/kured/pkg/delaytick"
+	"github.com/kubereboot/kured/pkg/leaselock"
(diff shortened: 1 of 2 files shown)


evrardjp:
I would like to dig deeper into this. I need to understand more about leases behaviour, api impact on large clusters. Okay to discuss this ? I added you on linkedin for better coordination.

PR #852, opened 2023-11-14
feat: add drain delay

MerlinDMC:
This PR adds a `--drain-delay` flag.
The duration given will delay the node draining process to allow for other controllers in the cluster to complete work.

We are using `--pre-reboot-node-labels=node.kubernetes.io/exclude-from-external-load-balancers=true` to signal the load balancer controller to start draining connections from the node as it is about to be drained.
With the flag added we can delay the draining process to ensure the load balancer controller has enough time to remove the node from active traffic flow before the draining starts.

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -44,6 +44,7 @@ var (
 
 	// Command line flags
 	forceReboot                     bool
+	drainDelay                      time.Duration
 	drainTimeout                    time.Duration
 	rebootDelay                     time.Duration
 	period                          time.Duration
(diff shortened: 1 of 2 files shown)


ckotzbauer:
Sometimes single test-run's are failing. This is unrelated to your changes.

ckotzbauer (approved):
Can you please open additional PRs to the website and chart-repo?
Thanks for the addition!!

MerlinDMC:
I've opened two additional PR to add the flag to the configuration properties and parameter docs.

ckotzbauer (approved):

jackfrancis (approved):
/lgtm

PR #873, opened 2024-01-04
feat: added alertmanager client package

atighineanu:
Hi everyone.

To finish this PR, I'd need more input on what kind of features for the alert manager client would you like to have. In the alert manager documentation one can set up tls, authentication. I've created an empty http.Client and would need more info about:
 - any token authentication for alert manager ?
 - tls ? e.g. https scheme when calling the api
 - additional flags? at least the ones I though of would be: alert manager url, token(if needed), http/https
 
 The client does (so far) only two things: 
(tested on an alert manager inside a kind cluster)
 Status().Get()
Screenshot 2024-01-04 at 10 05 40

 Silences().Get()
Screenshot 2024-01-04 at 10 05 59

Thanks

### DIFF
--- a/pkg/alerts/alertmanager/client.go
+++ b/pkg/alerts/alertmanager/client.go
@@ -68,6 +68,7 @@ func (s *StatusEndpoint) Get() (*StatusResponse, error) {
 	if err != nil {
 		return nil, err
 	}
+	request.Header.Add("Authentication", fmt.Sprintf("Bearer %s", s.Token))
 	response, err := s.Client.Client.Do(request)
 	if err != nil {
 		return nil, err
@@ -103,6 +104,7 @@ func (s *SilencesEndpoint) Get() ([]GettableSilence, error) {
 	if err != nil {
 		return nil, err
 	}
+	request.Header.Add("Authentication", fmt.Sprintf("Bearer %s", s.Token))
 	response, err := s.Client.Client.Do(request)
 	if err != nil {
 		return nil, err
--- a/pkg/alerts/alertmanager/client.go
+++ b/pkg/alerts/alertmanager/client.go
@@ -1,0 +1,119 @@
+package alertmanager
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"path"
+	"time"
+)
+
+const (
+	alertManagerPathPrefix = "/api/v2"
+	// the default context timeout for alert manager client
+	// feel free to change this value/set a corresponding env var if needed
+	defaultTimeOut = 10 * time.Second
+)
+
+// New is a constructor of AlertManagerClient
+//
+// if no url flag is given => error
+func New(alertManagerURL, alertManagerToken string) (*Client, error) {
+	if alertManagerURL == "" {
+		return nil, fmt.Errorf("no alert manager url found")
+	}
+	return &Client{
+		Token:   alertManagerToken,
+		HostURL: alertManagerURL,
+		Client:  new(http.Client),
+	}, nil
+}
+
+// Status builds the Status endpoint
+func (c *Client) Status() *StatusEndpoint {
+	return &StatusEndpoint{
+		Client: *c,
+	}
+}
+
+// Silences builds the Silences endpoint
+func (c *Client) Silences() *SilencesEndpoint {
+	return &SilencesEndpoint{
+		Client: *c,
+	}
+}
+
+// BuildURL builds the full URL for Status Endpoint
+func (s *StatusEndpoint) BuildURL() error {
+	url, err := url.Parse(s.HostURL)
+	if err != nil {
+		return err
+	}
+	url.Path = path.Join(alertManagerPathPrefix, "status")
+	s.FullURL = url.String()
+	return nil
+}
+
+// Get receives information about alert manager overall status
+func (s *StatusEndpoint) Get() (*StatusResponse, error) {
+	err := s.BuildURL()
+	if err != nil {
+		return nil, err
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeOut)
+	defer cancel()
+	request, err := http.NewRequestWithContext(ctx, http.MethodGet, s.FullURL, nil)
+	if err != nil {
+		return nil, err
+	}
+	response, err := s.Client.Client.Do(request)
+	if err != nil {
+		return nil, err
+	}
+	responseObject := new(StatusResponse)
+	err = json.NewDecoder(response.Body).Decode(responseObject)
+	if err != nil {
+		return nil, err
+	}
+	return responseObject, nil
+}
+
+// BuildURL builds the full URL for silences Endpoint
+func (s *SilencesEndpoint) BuildURL() error {
+	url, err := url.Parse(s.HostURL)
+	if err != nil {
+		return err
+	}
+	url.Path = path.Join(alertManagerPathPrefix, "silences")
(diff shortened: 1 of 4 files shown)


pkg/alerts/alertmanager/client.go
```diff
@@ -0,0 +1,118 @@
...
+
+// New is a constructor of AlertManagerClient
+func New() (*Client, error) {
+	if os.Getenv("ALERTMGR_URL") == "" {
+		return nil, fmt.Errorf("no alert manager url found")
+	}
+	return &Client{
+		Token:   os.Getenv("ALERTMGR_TOKEN"),
+		HostURL: os.Getenv("ALERTMGR_URL"),
+		Client:  new(http.Client),
```
ckotzbauer, line 29: Please use cli-flags instead (they can be populated with env-variables as well)

ckotzbauer (commented):
Sorry for the long delay and thanks for your work.
I left some basic comments, In general I think this is a good start. Maybe we can cleanup some types, but thats not a priority topic.
Regarding your questions:
- The protocol is part of the URL now and I think that's fine, no need for additional configs
- I think we should offer basic-auth at the beginning, TLS-Auth can be added afterwards if needed.

pkg/alerts/alertmanager/client.go
```diff
@@ -0,0 +1,118 @@
...
+	request, err := http.NewRequestWithContext(ctx, http.MethodGet, s.FullURL, nil)
+	if err != nil {
+		return nil, err
+	}
+	response, err := s.Client.Client.Do(request)
+	if err != nil {
+		return nil, err
+	}
+	responseObject := new(StatusResponse)
+	err = json.NewDecoder(response.Body).Decode(responseObject)
```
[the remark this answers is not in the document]
atighineanu: Ehmmmm. responseObject IS already a pointer.
new(Type) returns a *Type.

Unless you want me to create a var that is of StatusResponse{} value and Decode into &responseObject...

pkg/alerts/alertmanager/client.go
atighineanu: Perfect, thanks for feedback. That's exactly what I needed to know!

pkg/alerts/alertmanager/client.go
```diff
@@ -0,0 +1,118 @@
...
+	responseObject := make([]GettableSilence, 0)
+	err = json.NewDecoder(response.Body).Decode(&responseObject)
+	if err != nil {
+		return nil, err
+	}
+	if err := ValidateStatus(responseObject); err != nil {
+		return nil, err
+	}
+	return responseObject, nil
+}
```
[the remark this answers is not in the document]
atighineanu: I wanted to systematize the URL/endpoint construction and their methods to be split among them. 
So that so, when you call client.Silences().Get() just like in k8s client-go: client.CoreV1().Pods....

pkg/alerts/alertmanager/client.go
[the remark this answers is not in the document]
ckotzbauer: ah okay, I wasn't aware of new `new(...)` function returning pointers. Fine for me.

pkg/alerts/alertmanager/client.go
atighineanu: Done

pkg/alerts/alertmanager/client.go
```diff
@@ -0,0 +1,121 @@
...
+		return err
+	}
+	url.Path = path.Join(alertManagerPathPrefix, "status")
+	s.FullURL = url.String()
+	return nil
+}
+
+// Get receives information about alert manager overall status
+func (s *StatusEndpoint) Get() (*StatusResponse, error) {
+	err := s.BuildURL()
```
evrardjp, line 61: Why does this BuildURL() method exist, while the StatusEndpoint data structure receives the FullURL (which I suppose contains `status`)? 

If FullURL does not intend to be the complete URL (status included) and only intends to be base URL, then I guess having this logic could make sense.

However, in that case, I think I would implement it using the standard go net/url instead, because we have a recent enough version of go. This allows us to remove BuildURL() completely.

For example, in statusendpoint get(), I would do url.JoinPath(s.baseUrl,"status") (see https://pkg.go.dev/net/url@go1.20#JoinPath).

This automatically parses the url and will output any errors of parsing: https://cs.opensource.google/go/go/+/refs/tags/go1.20:src/net/url/url.go;l=1262.

Hence the purpose of L61 of this file is still a one liner, yet it removes the whole BuildURL.

On top of that, I feel it is easier to grasp the Get() if you have all the details there.

pkg/alerts/alertmanager/client.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		return nil, err
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeOut)
+	defer cancel()
+	request, err := http.NewRequestWithContext(ctx, http.MethodGet, s.FullURL, nil)
+	if err != nil {
+		return nil, err
+	}
+	request.Header.Add("Authentication", fmt.Sprintf("Bearer %s", s.Token))
```
evrardjp, line 71: what if there is no auth? Do we still have to add the header?

For example if AlertManager is empty.

pkg/alerts/alertmanager/client.go
```diff
@@ -0,0 +1,121 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"path"
+	"time"
+)
+
+const (
```
evrardjp, line 13: Did you consider using the alertmanager cli go module instead?
Would it be more complex to maintain?

pkg/alerts/alertmanager/types.go
```diff
@@ -0,0 +1,95 @@
+package alertmanager
+
+import (
+	"fmt"
+	"net/http"
+)
+
+var (
+	silenceStates = map[string]bool{"expired": true, "active": true, "pending": true}
```
evrardjp, line 9: Maybe a comment or another data structure would help here, as the intent is not obvious.

pkg/alerts/alertmanager/types.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// StatusEndpoint is the status enpoint of the alert manager client
+type StatusEndpoint struct {
+	Client  `json:"alertmanagerClient" yaml:"alertmanagerClient"`
+	FullURL string `json:"fullUrl" yaml:"fullUrl"`
+}
+
+// SilencesEndpoint is the silences enpoint of the alert manager client
+type SilencesEndpoint struct {
```
evrardjp, line 26: Isn't that the same structure?

If you still wish to use a not complete url and keep the different data types (did you think of interfaces?), you could then use a variable in such structure for relative endpoint, which could be "status" for StatusEndpoint. This could be filled by default with tags. It would improve readability IMO, but less than including the whole mumbojumbo in the Get().

... On top of that, I prefer the endpoint being the full URL to be honest :) I avoid the temptation to guess.

evrardjp (commented):
Good to read you!

I think I miss a few bits about the intention behind this code, and the next steps.

Yet, allow me to comment a bit on the code I read, which might help improve the readability later.

pkg/alerts/alertmanager/client.go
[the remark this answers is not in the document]
evrardjp: Thanks for explaining that. It clarifies your goals with the structure. I think this should be in a comment in the code.

PR #899, opened 2024-02-15
Make nsenter a configurable option

bh185140:
Closes https://github.com/kubereboot/kured/issues/896


ckotzbauer (commented):
Thanks for the PR, I think I would prefer a flag if nsenter should be included or not. These changes here give us more flexibility, but it is also more complicated to configure it correctly.
@jackfrancis WDYT?

PR #908, opened 2024-03-04
Propagate Context

leonnicolas:
Propagate a context from the main function to wherever it is needed.

Use `github.com/oklog/run` run groups to handle the life cycle of the go
routines running the rebooter and metrics server.

Ref: https://github.com/kubereboot/kured/issues/234
and https://github.com/kubereboot/kured/pull/808

Signed-off-by: leonnicolas <leonloechner@gmx.de>

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -426,7 +426,6 @@ func (kb KubernetesBlockingChecker) isBlocked(ctx context.Context) bool {
 }
 
 func rebootBlocked(ctx context.Context, blockers ...RebootBlocker) bool {
-	// TODO: can we do that concurrently
 	for _, blocker := range blockers {
 		if blocker.isBlocked(ctx) {
 			return true
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math/rand"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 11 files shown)


evrardjp (commented):
I think it's a good step. It implements what we talked about in the fast, with a real explanation of why we use context.

Yet, I am not the best person to judge wheter this is _enough_ or whether it's the right context. (And I sadly don't have enough time to delve into it).

Would you mind explaining why it mattered for you? And why did you want to use background context, please? And why you used oklog?

Thank you.

leonnicolas:
> Would you mind explaining why it mattered for you? 

Passing down a context makes sure that when a caller wants to cancel an operation, some function down the line will not block until it returns. Instead it would probably return a context canceled. 
In practice, when kured received a SIGINT, it can cancel all outgoing requests and finish processing all incoming requests to the metrics server before it exits.

> And why did you want to use background context, please?

This is normally the context you use in the "root" of the application. https://pkg.go.dev/context#Background

> And why you used oklog?

At the moment kured starts 2 concurrent processes (go threads) in its main function and the metrics server that blocks the main thread forever.
If any of the go threads crash, we would continue to run the metrics server. No way to tell the other thread to exit as well and terminate the main thread (or try to restart the crashed thread). 
oklog's run package just provides one way to manage the life cycle of threads. We could probably use other packages as well. It was just my personal favorite because its API is so slim.

cmd/kured/main.go
```diff
@@ -421,22 +425,23 @@ func (kb KubernetesBlockingChecker) isBlocked() bool {
 	return false
 }
 
-func rebootBlocked(blockers ...RebootBlocker) bool {
+func rebootBlocked(ctx context.Context, blockers ...RebootBlocker) bool {
+	// TODO: can we do that concurrently
```
jackfrancis, line 429: So that we don't commit a `// TODO` comment, could you add an issue describing this optimization instead? (agree that this is out of scope for this PR)

thank you!

PR #926, opened 2024-04-30
update k8s deps

dholbach:
Still waiting for 1.30 kind image to test with.


evrardjp (commented):
Looks good on paper.

I should start helping on those kind images, but didn't get the chance so far.

ckotzbauer (approved):

PR #930, opened 2024-05-08
fix: make sure node is drained before rebooting when rebootDelay is g…

### LINKED ISSUE #929: Kured should verify if node is still drained when delaying reboot
Hey, 

we had a problem with two controllers draining/uncordon nodes. To make the reboot more graceful, we configured `rebootDelay: 120s`. During these 2 mins the other controller uncordened the node. Kured rebooted the not drained node afterwards which is not optimal :D
So Kured should verify after the delay and right before the reboot is issued, if the node is still drained or not.

sebastiangaiser:
…reater than 0

closes #929

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -788,6 +788,18 @@ func rebootAsRequired(nodeID string, booter reboot.Reboot, sentinelCommand []str
 		if rebootDelay > 0 {
 			log.Infof("Delaying reboot for %v", rebootDelay)
 			time.Sleep(rebootDelay)
+
+			log.Infof("Making sure %s is drained before reboot", node.GetName())
+			err = drain(client, node)
+			if err != nil {
+				if !forceReboot {
+					log.Errorf("Unable to cordon or drain %s: %v, will release lock and retry cordon and drain before rebooting when lock is next acquired", node.GetName(), err)
+					release(lock, concurrency > 1)
+					log.Infof("Performing a best-effort uncordon after failed cordon and drain")
+					uncordon(client, node)
+					continue
+				}
+			}
 		}
 
 		if notifyURL != "" {


jackfrancis:
I've added some concerns in the original issue #929. Let's definitely figure out how to solve for this scenario, but I'm not sure that a second drain operation after the retry wait is the right solution.

cmd/kured/main.go
```diff
@@ -788,6 +788,18 @@ func rebootAsRequired(nodeID string, booter reboot.Reboot, sentinelCommand []str
 		if rebootDelay > 0 {
 			log.Infof("Delaying reboot for %v", rebootDelay)
 			time.Sleep(rebootDelay)
+
```
evrardjp, line 791: I don't think it make sense to add this conditionally on reboot delay.

cmd/kured/main.go
```diff
@@ -788,6 +788,18 @@ func rebootAsRequired(nodeID string, booter reboot.Reboot, sentinelCommand []str
...
+			log.Infof("Making sure %s is drained before reboot", node.GetName())
+			err = drain(client, node)
```
evrardjp, line 793: I don' think we should trigger another drain. 

The state should be as _expected_ before reboot, but not triggering an extra drain. See @jackfrancis 's opinion on the issue.

evrardjp (changes requested):
I don't think it's wise to add this _as is_. Let's discuss on the issue first maybe?

evrardjp:
I like the idea, but I think we have details to iron out.

sebastiangaiser:
Not stale...

evrardjp:
@sebastiangaiser  maybe we could have a meeting and you explain your point of view/what's missing to progress this? I like the concept.

sebastiangaiser:
I'm sorry but I don't have the time, currently. Feel free to hijack/supersede my PR.

PR #961, opened 2024-07-29
Support addtional filter for extended resource pod

fishingfly:
This PR is related to issue https://github.com/kubereboot/kured/issues/960

# How to use
```diff
containers:
        - name: kured
          image: ghcr.io/kubereboot/kured:1.16.0
          command:
            - /usr/bin/kured
            - --reboot-sentinel=/sentinel/reboot-required
+          - --drain-resource-name-with-prefixes=nvidia.com/gpu,nvidia.com/mig
```
By using `drain-resource-name-with-prefixes=nvidia.com/gpu,nvidia.com/mig` flag, we can only evict pods with nvidia gpu resourece or mig device resource.

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -30,6 +30,7 @@ import (
 	"github.com/google/shlex"
 
 	shoutrrr "github.com/containrrr/shoutrrr"
+	"github.com/kubereboot/kured/pkg/additionalfilters"
 	"github.com/kubereboot/kured/pkg/alerts"
 	"github.com/kubereboot/kured/pkg/daemonsetlock"
 	"github.com/kubereboot/kured/pkg/delaytick"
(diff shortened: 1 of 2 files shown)


yeahdongcn:
This looks super useful for GPU clusters.

evrardjp:
So the current pod filter is completely not okay for you?

fishingfly:
> So the current pod filter is completely not okay for you?

Our scenario is in a gpu cluster and we only want to evict all gpu pods before the machine is restarted. The current pod filter is based on Label. If we use the current pod filter, the modification we need to make is: add a fixed label to all gpu pods. We do not want to modify the user's pods.

evrardjp:
"We do not want to modify the user's pods." That's the key phrase. Got it now.

I need to ponder a bit on the code now. Can't say _when_ I will be able to do this though, so I won't block any progress happening in the meantime.

I have no strong opinion about the fact it should be included or not. FYI, I am naturally leaning to not include new features for ease of maintenance in the long run (saying no is temporary, saying yes is definitive!) , especially if it's a feature can be circumvented by training users. But I understand your position and your case. If the code can be written in a way that's simple to maintain, that's all good.

maratrixx:
We also need this feature and look forward to accepting it soon. thx～

maratrixx:
Look forward to accepting it soon. thx ~

PR #971, opened 2024-09-03
970: Taint nodes even if reboot is currently blocked

### LINKED ISSUE #970: Allow tainting Nodes even when blocker Pods exist
I'm proposing an optimization that should help a cluster perform a full reboot faster when there are blocker conditions.

Currently, the [logic is as follows](https://github.com/kubereboot/kured/blob/main/cmd/kured/main.go#L764-L775) when using both `--blocking-pod-selector` and `--prefer-no-schedule-taint` :

```go
var rebootRequiredBlockCondition string
if rebootBlocked(blockCheckers...) {
	rebootRequiredBlockCondition = ", but blocked at this time"
	continue
}
log.Infof("Reboot required%s", rebootRequiredBlockCondition)

if !holding(lock, &nodeMeta, concurrency > 1) && !acquire(lock, &nodeMeta, TTL, concurrency) {
	// Prefer to not schedule pods onto this node to avoid draing the same pod multiple times.
	preferNoScheduleTaint.Enable()
	continue
}
```

By this point, we know that the Node requires a reboot, but if a blocker exists (e.g. Prometheus alert is firing or Pods exist on the node that match the blocking selector), the main loop will just go back to sleep and wait for the next tick without tainting the node. The problem here though is that more blocker Pods could be scheduled to this node while we wait for the next tick cycle to happen, meaning that it could take an extensively long period before the Node happens to be free of blocker Pods and can be rebooted.

My proposal is to add a flag so that Nodes will be tainted as `PreferNoSchedule` as soon as they're detected as requiring a reboot, then the block checker can continue as normal. This way, there is a high chance that blocker Pods will schedule on other nodes instead, as long as the scheduler can accommodate them somewhere else. Once all of the blocking conditions have cleared, the Node will reboot as normal.

I'm happy to submit a PR to create this flag if it's agreed to!

yaraskm:
- If using the --prefer-no-schedule-taint flag, apply the taint even if the reboot is currently blocked. This will make it less likely for Pods to get scheduled to nodes that are pending a reboot, once the blocking Pods / alerts are cleared.

Resolves #970

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -761,16 +761,23 @@ func rebootAsRequired(nodeID string, booter reboot.Reboot, sentinelCommand []str
 			blockCheckers = append(blockCheckers, KubernetesBlockingChecker{client: client, nodename: nodeID, filter: podSelectors})
 		}
 
+		rebootBlocked := rebootBlocked(blockCheckers...)
+
 		var rebootRequiredBlockCondition string
-		if rebootBlocked(blockCheckers...) {
+		if rebootBlocked {
 			rebootRequiredBlockCondition = ", but blocked at this time"
-			continue
 		}
+
 		log.Infof("Reboot required%s", rebootRequiredBlockCondition)
 
 		if !holding(lock, &nodeMeta, concurrency > 1) && !acquire(lock, &nodeMeta, TTL, concurrency) {
 			// Prefer to not schedule pods onto this node to avoid draing the same pod multiple times.
 			preferNoScheduleTaint.Enable()
+			continue
+		}
+
+		if rebootBlocked {
+			// We've logged that the reboot is needed, but curently blocked, and have tainted the node.
 			continue
 		}
 
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -770,14 +770,15 @@ func rebootAsRequired(nodeID string, booter reboot.Reboot, sentinelCommand []str
 
 		log.Infof("Reboot required%s", rebootRequiredBlockCondition)
 
-		if !holding(lock, &nodeMeta, concurrency > 1) && !acquire(lock, &nodeMeta, TTL, concurrency) {
-			// Prefer to not schedule pods onto this node to avoid draing the same pod multiple times.
+		if rebootBlocked {
+			// Prefer to not schedule pods onto this node to avoid draining the same pod multiple times.
 			preferNoScheduleTaint.Enable()
+			// We've logged that the reboot is needed, but curently blocked, and have tainted the node.
 			continue
 		}
 
-		if rebootBlocked {
-			// We've logged that the reboot is needed, but curently blocked, and have tainted the node.
+		if !holding(lock, &nodeMeta, concurrency > 1) && !acquire(lock, &nodeMeta, TTL, concurrency) {
+			// If we can't acquire the lock, poll again
 			continue
 		}
 


cmd/kured/main.go
```diff
@@ -761,16 +761,24 @@ func rebootAsRequired(nodeID string, booter reboot.Reboot, sentinelCommand []str
 			blockCheckers = append(blockCheckers, KubernetesBlockingChecker{client: client, nodename: nodeID, filter: podSelectors})
 		}
 
+		rebootBlocked := rebootBlocked(blockCheckers...)
```
jackfrancis, line 764: I wonder is the easiest and most correct way to improve this is to move right after L32 above, immediately after we determine that the node will need to be deleted.

As far as I can see there isn't any harm in *always* adding this "prefer no schedule" taint. Right now we're only doing it if we aren't able to rapidly progress to the drain operation, but why not do it no matter what?

This would solve for your current problem as well.

Thoughts?

yaraskm: Sounds good to me, I'll make the change later today.

yaraskm: Hey @jackfrancis , which line is it you're referring to? I don't think L32 is correct.

I've been able to successfully run with this PR in prod for the last few weeks now and it definitely seems to have lowered the average time for a node to be blocked. I'm definitely open to adding the taint all of the time, regardless of whether we're in an acceptable reboot window or not.

evrardjp:
I am not convinced this is a view everyone will have.

I believe some ppl will prefer adding the preferNoSchedule taint, while some others will consider that the taint shouldn't be liberally applied on all nodes including the ones that can't reboot.

I don't like the fact to change the rebootasrequired loop through a variable. It makes it very hard to understand the loop in terms of maintenance in the long run. On top of that, most of the reasons ppl would have adjusted the rebootasrequired over time was for personal preference: "In my case, I would like to do x before y". It all made sense for some ppl, while it didn't make sense for some others.

I truely belive your case makes sense. Yet, I am not certain that it's the case for everyone.

Two choices: 
- I will contact a few ppl, see what they think.
- We start to work on a change that makes the rebootasrequired loop more flexible.

cmd/kured/main.go
evrardjp: @jackfrancis  do you truly believe that adding the prefer no schedule taint can be done in all cases?
Wouldn't it make it a bit moot to have all nodes with prefernoschedule after a security update (for which all nodes will get the taint)? When kubernetes will schedule (all the nodes having the prefernoschedule) there will be no distinction anymore... Am I missing something?

yaraskm:
> I truely belive your case makes sense. Yet, I am not certain that it's the case for everyone.

I definitely agree with you here, and making it a configurable behaviour change makes the most sense to me as well. I'll wait for your feedback on the discussions and am happy to chip in on implementation changes if needed.

nikokirilenko:
> I truely belive your case makes sense. Yet, I am not certain that it's the case for everyone.


We are exploring using kured in our k8s environment which is hosting a Jenkins infrastructure, meaning the Jenkins controller spins up ephemeral pods acting as Jenkins agent which run various CI pipelines.
We can't evict those Jenkins pods as killing off the agent pod basically kills the Jenkins build.
For our use-case we would like to:

1. Taint the node to prevent any further Jenkins pods scheduling on it. (`--prefer-no-schedule-taint=weave.works/kured-node-reboot`)
2. Wait for all Jenkins pods to finish their workload. (`--blocking-pod-selector=jenkins=agent`)
3. Continue with reboot etc...

This PR will make this use-case feasible and will be much appreciated.

evrardjp:
Can jenkins apply its own taint when informed about a node condition? 

In v2 process, it will be quite easier to have your own blocker controller that watches jenkins pods and applies taint + condition when kured or another node maintenance system applies a "maintenance" condition to the node. 

The reboot will only be triggered when that inhibitor is cleaning the inhibiting condition. It's very clean this way. 
Code is really jenkins specific and is about two hundred lines... But you will be able to do an event based system that will react exactly how you want, instead of making kured what it isn't.

I am okay to discuss this in a community meeting if you wish.

PR #981, opened 2024-10-01
Fix ci

evrardjp:


evrardjp:
All tests passed without issues in 6minutes and a half. I think that's better than what we had :)

jackfrancis (approved):
This all seems sane, great chop 🪵 carry 🪣 vibes here.

lgtm

PR #982, opened 2024-10-06
Cleanup Code and CI

evrardjp:
Without this, it is hard to adapt the code and test it locally.

This MR contain multiple changes, separately mergeable. In order to faster iterate, I made a single PR.

With this MR, we can more easily test locally kured, and the CI is closer to the local tests.

The next step, after this merges, is to bump golang, kube versions, and continue cleaning up the code.

This will open the way for the following PRs:
- A refactoring of the locks
- A refactoring of the flag handling to simplify it and make it closer to a pure function.
- A cleanup of rebootAsRequired to make its workflow more readable.


evrardjp:
Please do not squash merge this PR.

evrardjp:
Please do not merge yet.

evrardjp:
Bonus feature added on CI: Testing that cordoning before the reboot stays applied.

evrardjp:
Failure due to rate limiting of trivy :/

evrardjp:
Updated the CI matrix to make it a bit simpler.

evrardjp:
fixed conflicts due to recent merges of .github worklows.

Makefile
```diff
@@ -14,10 +14,12 @@ $(TEMPDIR):
 
 .PHONY: bootstrap-tools
 bootstrap-tools: $(TEMPDIR)
-	VERSION=v1.24.0 TMPDIR=.tmp bash .github/scripts/goreleaser-install.sh
-	curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b .tmp v1.0.1
-	curl -sSfL https://github.com/sigstore/cosign/releases/download/v2.2.3/cosign-linux-amd64 -o .tmp/cosign
-	chmod +x .tmp/goreleaser .tmp/cosign .tmp/syft
+	command -v .tmp/goreleaser || VERSION=v1.24.0 TMPDIR=.tmp bash .github/scripts/goreleaser-install.sh
```
jackfrancis, line 17: I know you just inherited this but I think the relative directory `hack/bin/` or `hack/tools/bin` is more idiomatic than `.tmp/` for installing and running supporting binaries at runtime. Maybe a follow-up?

Makefile
```diff
@@ -35,7 +37,23 @@ kured-release-snapshot:
 	$(GORELEASER_CMD) release --clean --snapshot
 
 image: kured
-	$(SUDO) docker buildx build --load -t ghcr.io/$(DH_ORG)/kured:$(VERSION) .
+	$(SUDO) docker buildx build --no-cache --load -t ghcr.io/$(DH_ORG)/kured:$(VERSION) .
+
+dev-image: image
+	$(SUDO) docker tag ghcr.io/$(DH_ORG)/kured:$(VERSION) kured:latest
```
jackfrancis, line 43: would `canary` be a better tag than `latest` for an ephemeral dev image?

Makefile
```diff
@@ -35,7 +37,23 @@ kured-release-snapshot:
...
+
+dev-manifest:
+	# basic e2e scenario
+	sed -e "s#image: ghcr.io/.*kured.*#image: kured:latest#g" -e 's/#\(.*\)--period=1h/\1--period=20s/g' kured-ds.yaml > tests/kind/testfiles/kured-ds.yaml
```
jackfrancis, line 47: if we change the `latest` tag remember to change here as well :)

jackfrancis:
Overall this looks good. The safest way to land this would probably be to cleanup the CI first as a separate PR (and then potentially iterate a bit if we discover gremlins), and then when we are very confident in the health of our CI and tests, we can land the refactor, which *should just pass the E2E* tests (my first read of the code suggests that there are no [intended] functional changes to the behavior, just code optimization).

Great work, this will definitely be way easier to maintain! :)

Makefile
evrardjp: I would love to do it either in a follow up or here, whatever merges first :)

Makefile
evrardjp: I am fine with that.

evrardjp:
There is a small hiccup with the approach of adding CI first, is that right now we have bugs, and the CI caught it ;) 

The refactor is fixing some of those. However, I am fine to have that MR being just the CI, we merge with our code being broken, and we refactor to get to the better place. This way it makes it very clear of the intent.

Makefile
evrardjp: I went to "dev" instead of canary, as I think it makes it more explicit that it's not yet ready for release :)

Makefile
evrardjp: done in this MR.

Makefile
```diff
@@ -45,10 +63,9 @@ manifest:
...
-	go test ./...
-	echo "Running golint on pkg"
-	golint ./pkg/...
-	echo "Running golint on cmd"
-	golint ./cmd/...
+test: bootstrap-tools
+	echo "Running short go tests"
+	go test -test.short -json ./... > test.json
+	echo "Running shellcheck"
+	find . -name '*.sh' -exec .tmp/shellcheck  {} \;
```
dholbach, line 70: Path needs to be adjusted.

dholbach: It doesn't seem to fail either: https://github.com/kubereboot/kured/actions/runs/11305350571/job/31444749412?pr=982#step:4:116

Makefile
```diff
@@ -1,23 +1,25 @@
 .DEFAULT: all
 .PHONY: all clean image minikube-publish manifest test kured-all
 
-TEMPDIR=./.tmp
```
dholbach, line 4: `.tmp` seems to be still used in other places.

Makefile
evrardjp: I am not a Makefile expert, this behaviour was already present. Any idea?

Makefile
evrardjp: I didn't find it in another places. The grep didn't return any result. What else did I miss?

evrardjp:
@dholbach  I removed the .tmp in the find.

jackfrancis:
@evrardjp this looks like just CI changes now, correct?

/lgtm

evrardjp:
@jackfrancis  can you approve then?

jackfrancis (approved):
/approve

evrardjp:
Sadly this was merged into a single commit, so I hope we won't need a partial revert. I will keep the branch remove_flags temporarily, in case we need a revert.

Issue #987, opened 2024-10-11
Update k8s deps to the latest 1.29


dholbach:
Maybe update k8s deps to the latest as well with it? https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.29.md#changelog-since-v1298

_Originally posted by @dholbach in https://github.com/kubereboot/kured/issues/978#issuecomment-2345390707_

PR #989, opened 2024-10-11
Update k8s deps and images

dholbach:
Fixes: #987


evrardjp:
If we merged https://github.com/kubereboot/kured/pull/982 first, it would be easier to test this PR... Could you think about it? I have reduced its scope, and it should be just enough to test this MR efficiently.

evrardjp (approved):
All tests passed locally and in CI.
I know I pushed the last commit, but it's basically a simple rebase on dholbach's fork to make it work. Therefore I don't consider myself as author, and can accept the code as is.

PR #990, opened 2024-10-12
Cleanup Part 1 - First refactors

evrardjp:
This contains a few commits there were
historically in the MR 982 [1], to refactor
kured for more maintainability.

This will be the first in a series of
refactors.

[1]: https://github.com/kubereboot/kured/pull/982


evrardjp:
This could be merged _after_ PR 989 if you want.

evrardjp:
DCO issue on a commit, pushed it back. 
I rebased on top of the changes asked in the PR for CI.

evrardjp:
I don't like the idea to create so many small packages, with exported variables. Yet, I find that this is the best way to achieve the goals of cleanup. I intend to continue cleaning up with a part2, doing larger changes, then internalizing many packages.

pkg/reboot/command.go
```diff
@@ -1,25 +1,32 @@
...
-	log.Infof("Running command: %s for node: %s", c.rebootCommand, c.nodeID)
-	if err := util.NewCommand(c.rebootCommand[0], c.rebootCommand[1:]...).Run(); err != nil {
-		log.Fatalf("Error invoking reboot command: %v", err)
+// NewCommandRebooter is the constructor to create a CommandRebooter from a string not
+// yet shell lexed. You can skip this constructor if you parse the data correctly first
+// when instantiating a CommandRebooter instance.
+func NewCommandRebooter(rebootCommand string) *CommandRebooter {
+	cmd, err := shlex.Split(rebootCommand)
+	if err != nil {
+		log.Fatalf("Error parsing provided reboot command: %v", err)
```
jackfrancis, line 28: Maybe output the `rebootCommand` string here in case the error response from `shlex` doesn't include it

pkg/reboot/command.go
```diff
@@ -1,25 +1,32 @@
...
+type CommandRebooter struct {
+	RebootCommand []string
 }
 
-// NewCommandReboot creates a new command-rebooter which needs full privileges on the host.
-func NewCommandReboot(nodeID string, rebootCommand []string) *CommandRebootMethod {
-	return &CommandRebootMethod{nodeID: nodeID, rebootCommand: rebootCommand}
+// Reboot triggers the reboot command
+func (c CommandRebooter) Reboot() {
+	log.Infof("Invoking command: %s", c.RebootCommand)
```
jackfrancis, line 16: I don't know how this would ever occur but because `RebootCommand` is an `[]string` there could be a nil panic here if `RebootCommand` is nil

pkg/reboot/command.go
evrardjp: that reminds me that I should remove the fatalf from there :)

pkg/daemonsetlock/daemonsetlock.go
```diff
@@ -28,25 +45,92 @@ type DaemonSetLock struct {
...
+			},
+		}
+	}
+}
+
+// GetDaemonSet returns the named DaemonSet resource from the DaemonSetLock's configured client
+func (dsl *DaemonSetLock) GetDaemonSet(sleep, timeout time.Duration) (*v1.DaemonSet, error) {
+	var ds *v1.DaemonSet
+	var lastError error
+	err := wait.PollImmediate(sleep, timeout, func() (bool, error) {
```
jackfrancis, line 118: `PollUntilContextTimeout` might be cleaner here

https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollUntilContextTimeout

pkg/reboot/command.go
evrardjp: I will ensure that in the constructor.

pkg/checkers/checker.go
```diff
@@ -0,0 +1,98 @@
...
+	"os"
+	"os/exec"
+)
+
+// Checker is the standard interface to use to check
+// if a reboot is required. Its types must implement a
+// CheckRebootRequired method which returns a single boolean
+// clarifying whether a reboot is expected or not.
+type Checker interface {
+	CheckRebootRequired() bool
```
jackfrancis, line 16: `CheckRebootRequired()` as a method for `Checker` has a bit of a stutter, maybe just `RebootRequired()` is sufficient

cmd/kured/pflags.go
```diff
@@ -1,31 +1,34 @@
...
 type regexpValue struct {
 	value **regexp.Regexp
 }
 
+// String method to return the string representation of the value
 func (rev *regexpValue) String() string {
-	if *rev.value == nil {
-		return ""
+	if *rev.value != nil {
+		return (*rev.value).String()
```
jackfrancis, line 16: ya this is better

cmd/kured/pflags.go
```diff
@@ -1,31 +1,34 @@
 package main
 
 import (
+	"fmt"
 	"regexp"
 )
 
+// Custom flag type to handle *regexp.Regexp with pflag
 type regexpValue struct {
 	value **regexp.Regexp
```
jackfrancis, line 10: is this a missed opportunity to reconsider our "pointer to a pointer" approach here? :)

evrardjp: Totally. Probably warrant having a looksie later tomorrow. Tonight is time for bed.

FYI I tried my best to keep this MR as "refactor only", and do slightly bigger changes with possible behaviour changes in the other PR.

cmd/kured/main.go
```diff
@@ -121,375 +113,302 @@ func init() {
...
+	log.Infof("Reboot method: %s", rebootMethod)
+	var rebooter reboot.Rebooter
+	switch {
+	case rebootMethod == "command":
+		log.Infof("Reboot command: %s", rebootCommand)
+		rebooter = reboot.NewCommandRebooter(rebootCommand)
+	case rebootMethod == "signal":
+		log.Infof("Reboot signal: %v", rebootSignal)
+		rebooter = reboot.NewSignalRebooter(rebootSignal)
+	default:
```
jackfrancis, line 255: Can we move this invalid configuration case into the input validation flow so we can assume by the time we get here we don't have to worry about it?

evrardjp: WDYM? I do input validation into each constructor. Do you want to have a reboot.NewRebooter, which will take the rebootMethods?

pkg/daemonsetlock/daemonsetlock.go
evrardjp: Will do on follow up commit.

cmd/kured/main.go
evrardjp: Let's see if the latest patch pleases you :)

evrardjp:
@jackfrancis  can you please not squash merge this time? If necessary, I will adapt the commits to make it look cleaner, but I really dislike to lose history, especially when those commits are mostly atomic.

cmd/kured/pflags.go
evrardjp: I pushed a new version including this fix. I would love it to stay as a separate commit.

pkg/daemonsetlock/daemonsetlock.go
evrardjp: Done in a new commit, but not really confident here, as it has no test coverage. Can you double check on your side?

evrardjp:
I added a few extra commits, to merely manage the API surface, and remove deps. I will stop here, it's large enough.

evrardjp:
@ckotzbauer I know you're quite busy nowadays... Could you tell us if you have time to review this/give your opinion?
Else I will just continue with @jackfrancis .

evrardjp:
@jackfrancis should we get the ball rolling?

pkg/daemonsetlock/daemonsetlock.go
jackfrancis: looks sane to me

cmd/kured/main.go
jackfrancis: I think this is better!

jackfrancis (approved):
lgtm

jackfrancis:
@evrardjp feel free to merge w/ your preferred commit preservation strategy

evrardjp:
Thanks @jackfrancis  for the reviews.

PR #995, opened 2024-10-17
Clarify features for e2e test

evrardjp:
Without this, ppl might be wondering how to use e2e test, especially on slower machines.


jackfrancis (approved):
lgtm

PR #996, opened 2024-10-17
Cleanup dockerfile

evrardjp:
Without this, docker buildx will complain about the
following issues:
- "RedundantTargetPlatform: Setting platform to predefined $TARGETPLATFORM in
   FROM is redundant as this is the default behavior"
- "FromAsCasing: 'as' and 'FROM' keywords' casing do not match"

This fixes it.


jackfrancis (approved):
lgtm

PR #1000, opened 2024-10-18
[WIP] Cleanup Part 2 - Rewrite

evrardjp:
This is the follow up PR to https://github.com/kubereboot/kured/pull/990 .
It tackles deeper changes to reach a 2.0 of the rewrite.

What it does:
- The main goroutine, cleaning up the code for easier maintenance
- Remove the usage of logrus, as we can simply use slog
- Makes the system more unit testable.

Still to do (in another PR):
- Ensure logs are out for all the loops.
- Be generic on the evacuator
- The split into different components for v2
- Take a stance on simplification (remove lock TTLs, annotate by default)


evrardjp:
I need to reimplement this from scratch, please ignore my changes from now, until I mention this is ready for review (and [WIP] flag will be removed)

evrardjp:
should I bother revive this, @jackfrancis ?
This was the place for large refactoring/cleaning of the code, to be able to replace some implementations by others.
The intent was to be able to use another notifyer (replace shoutrrr), or implement different lock mechanism.
With this in mind, I completely cleaned up the lock code, fixing a long standing security point, but at the expense of breaking the current behaviour. 

I am fine to resume this work, but only if we'll eventually merge it. No point for me to always rebase this massive change if it doesn't get in.

evrardjp:
I will need to reopen this and work on it, after a conversation about our direction.

Issue #1013, opened 2024-11-02
Leaking goroutines


evrardjp:
This is unrelated to #1004, as this also appears in 1.16.0.

evrardjp:
Apparently comes from our usage of logrus. Again, I got a branch fixing that, but I need to bring another solution that doesn't change the main loop. If anyone has a solution (I don't have much time to look at this rn), please feel free to propose a patch.

jonaz:
It does not leak in 1.16.0 for us. We rolled back to that since .1 leaks both goroutines and memory:

here is a graph of go_goroutines{pod=~"kured.*"} showing the gooutine leak. The other issue already have memory graph. 

![Image](https://github.com/user-attachments/assets/b13897e3-369b-49b8-8e2d-347f7e1132d7)

jonaz:
The leak is from every time NewCommand i called you leak goroutine in logrus.WriterLevel ![Image](https://github.com/user-attachments/assets/eee93316-4d70-4324-a0da-545af4824129)

jonaz:
You want me to make a PR with the fix?

evrardjp:
> The leak is from every time NewCommand is called you leak goroutine in logrus.WriterLevel ![Image](https://github.com/user-attachments/assets/eee93316-4d70-4324-a0da-545af4824129)
> 
> This might also be related to the memoryleak.

From my quick glance a few days ago, that's EXACTLY what I understood.
In my fork, I am replacing quite a few calls with slog. The refactor is a bit big, so I prefer NOT bringing this refactor in _as is_, and I prefer remplementing a fix.

I am fine with you committing a PR, it will get us to land this quicker, IMO. When do you think you can propose this? I am available for a bit in the next hours, which should be enough to fix this with your help.

In terms of the memory leak, my guess was that it's not _completely_ related. (The fact that we also append in the rebootAsRequired loop is crazy, and is also worth a fix).

Yet, IF the memory leak appeared in 1.16.1 (and a revert of 1.16.0 fixes the issue), this needs further investigation: Logrus didn't change. We changed go version and prom client. Go version could have an impact.
My guess is that we fix the goroutine leak first, then work on the rebootAsRequired (I would love a review if you're okay), then with thie cleaned up, we can find other big items.

Can you confirm that 1.16.0 is NOT having a memory leak for you @jonaz  ?

jonaz:
@evrardjp yes 1.16.0 have no memory or goroutine leaks. It might be related to go version yes.

[...]

Issue #1004, opened 2024-10-25
Constantly increasing memory usage since upgrade to 1.16.1


ggogel:
I'm running a Kubernetes Cluster using the Terraform module kube-hetzner. It uses kured for automatic reboots. Since the upgrade to version 1.16.1, I've noticed that the memory usage of the kured pods is constantly increasing. This looks like a memory leak to me. 

![Image](https://github.com/user-attachments/assets/275e285a-518e-4d67-a8e5-073ec60781af)

K8s version: `v1.30.5+k3s1`

Output of `kubectl logs -l name=kured -n kube-system`:
```
time="2024-10-23T21:17:56Z" level=info msg="Lock Annotation: kube-system/kured:weave.works/kured-node-lock"
time="2024-10-23T21:17:56Z" level=info msg="Lock TTL not set, lock will remain until being released"
time="2024-10-23T21:17:56Z" level=info msg="Lock release delay not set, lock will be released immediately after rebooting"
time="2024-10-23T21:17:56Z" level=info msg="PreferNoSchedule taint: "
time="2024-10-23T21:17:56Z" level=info msg="Blocking Pod Selectors: []"
time="2024-10-23T21:17:56Z" level=info msg="Reboot schedule: Sun------------------ between 03:00 and 08:00 UTC"
time="2024-10-23T21:17:56Z" level=info msg="Reboot check command: [test -f /sentinel/reboot-required] every 5m0s"
time="2024-10-23T21:17:56Z" level=info msg="Concurrency: 1"
time="2024-10-23T21:17:56Z" level=info msg="Reboot method: command"
time="2024-10-23T21:17:56Z" level=info msg="Reboot signal: 39"
time="2024-10-23T21:12:40Z" level=info msg="Lock Annotation: kube-system/kured:weave.works/kured-node-lock"
time="2024-10-23T21:12:40Z" level=info msg="Lock TTL not set, lock will remain until being released"
time="2024-10-23T21:12:40Z" level=info msg="Lock release delay not set, lock will be released immediately after rebooting"
time="2024-10-23T21:12:40Z" level=info msg="PreferNoSchedule taint: "
time="2024-10-23T21:12:40Z" level=info msg="Blocking Pod Selectors: []"
time="2024-10-23T21:12:40Z" level=info msg="Reboot schedule: Sun------------------ between 03:00 and 08:00 UTC"
[...]

PR #1015, opened 2024-11-05
Fix goroutine leak

evrardjp:
Without this patch, we use WriterLevel, which spawns go routines. As we do it at every call of the util commands, we spawn goroutines at every check.

This is a problem as it leads to memory management issues.

This fixes it by using a buffer for stdout and stderr, then logging the results after the command was executed.

To make sure the logging happened at the same place, I inlined the code from utils. This results in duplicated the code.

However, this is not a big problem as:
- It makes the code more readable
- The implementation between checkers and rebooters _ARE_ different -- One definitely NEEDS privileges, while the other does not... Which could lead to later improvements.

Removing a "utils" package is not really a big deal (it is kinda a win in itself, as it is an anti-pattern), as the test coverage was kept.

Partial-Fix: #1004
Fixes: #1013


evrardjp:
@jonaz Could you check this, please?

pkg/checkers/checker.go
```diff
@@ -53,16 +53,15 @@ type CommandChecker struct {
...
-	cmd := util.NewCommand(cmdline[0], cmdline[1:]...)
+	bufStdout := new(bytes.Buffer)
+	bufStderr := new(bytes.Buffer)
+	cmd := exec.Command(rc.CheckCommand[0], rc.CheckCommand[1:]...)
+	cmd.Stdout = bufStdout
+	cmd.Stderr = bufStderr
+
 	if err := cmd.Run(); err != nil {
 		switch err := err.(type) {
 		case *exec.ExitError:
```
jonaz, line 67: We might want to output stderr and stdout in the log.Warnf below so the user gets important output. Previously this was displayed with the io.Pipe in logrus writer.

pkg/reboot/command.go
```diff
@@ -15,9 +16,17 @@ type CommandRebooter struct {
...
-	if err := util.NewCommand(c.RebootCommand[0], c.RebootCommand[1:]...).Run(); err != nil {
+
+	bufStdout := new(bytes.Buffer)
+	bufStderr := new(bytes.Buffer)
+	cmd := exec.Command(c.RebootCommand[0], c.RebootCommand[1:]...)
+	cmd.Stdout = bufStdout
+	cmd.Stderr = bufStderr
+
+	if err := cmd.Run(); err != nil {
 		return fmt.Errorf("error invoking reboot command %s: %v", c.RebootCommand, err)
```
jonaz, line 27: We do want to add both stderr and stdout to the error returned. Could contain important info about the error.

evrardjp: sure.

pkg/checkers/checker.go
evrardjp: fine for me.

evrardjp: Side note, when error happens, most likely the stderr and stdout will be empty. Do we really need this?

evrardjp:
@jonaz  can you have a second look please?

pkg/checkers/checker.go
jonaz: it totally depends on the command and hiding the error is not helpful for the end user.

evrardjp:
I tested commit 1 on a new env. It has good results.

jonaz:
@evrardjp LGTM :+1:

jonaz (approved):

dholbach (approved):

PR #1016, opened 2024-11-05
Cleanup blockCheckers

evrardjp:
Do not merge until #1015  is merged.

There is no need to continuously reallocate the check blockers.
They only need to be defined once.

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -262,6 +262,13 @@ func main() {
 		log.Fatal(err)
 	}
 
+	var blockCheckers []blockers.RebootBlocker
+	if prometheusURL != "" {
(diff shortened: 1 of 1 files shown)


cmd/kured/main.go
```diff
@@ -645,14 +652,6 @@ func rebootAsRequired(nodeID string, rebooter reboot.Rebooter, checker checkers.
 			}
 		}
 
-		var blockCheckers []blockers.RebootBlocker
```
jonaz, line 648: I dont think this was a leak. Have you confirmed with pprof? ( i added PR for enable pprof through config)

since blockCheckers is a new var on every loop the old one should be GCed. 

But the new code is still better so i would keep it as in the PR :)

evrardjp: It is NOT a leak. but it annoys my brain. And I don't know if could be a future leak. I will change the commit message.

dholbach (approved):

PR #1017, opened 2024-11-06
Add pprof endpoints if enabled in configuration

jonaz:
To help troubleshoot future leaks in production.

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -48,6 +48,7 @@ var (
 	period                          time.Duration
 	metricsHost                     string
 	metricsPort                     int
+	pprofEnabled                    bool
 	drainGracePeriod                int
 	drainPodSelector                string
 	skipWaitForDeleteTimeoutSeconds int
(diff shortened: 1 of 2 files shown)


evrardjp:
I guess that's better than rebuilding... Ok will test it. The GET feels a bit awkward compared to using mux handler, but that's definitely a nit. Let me test that.

jonaz:
> I guess that's better than rebuilding... Ok will test it. The GET feels a bit awkward compared to using mux handler, but that's definitely a nit. Let me test that.

The method (GET) pattern is new in 1.21 i think it was. Docs: https://pkg.go.dev/net/http#ServeMux

cmd/kured/main.go
```diff
@@ -278,6 +281,9 @@ func main() {
 	go rebootAsRequired(nodeID, rebooter, rebootChecker, window, lock, client)
 	go maintainRebootRequiredMetric(nodeID, rebootChecker)
 
+	if pprofEnabled {
```
evrardjp, line 284: I don't think this works: while pprofEnabled is by defaut to false, the net/http/pprof is still loaded, so it's still injecting itself?

evrardjp (changes requested):
If it's always enabled, we don't need the handleFunc. Are you sure this can be disabled, or did I test wrong?

cmd/kured/main.go
jonaz: Oh  yeah.. :facepalm: 

i guess we could just `import _ "net/http/pprof"` and leave it enabled. 
Or use a custom mux and not the DefaultServeMux where we can conditionally enable it. 

Which do you like?

evrardjp: I don't know yet... Do we really need to have this enabled at all times? For me we do not... And we do not have to flip the switch live on a build. I don't mind continuing to add `import _ "net/http/pprof"` line when necessary.

For me the valid question is not about whether we do have profiling, but rather "who is checking the profiling before release" :)

evrardjp: let's just add `import _ "net/http/pprof"` then?

jonaz: Yeah that's why I would like it default off but on for all our hundreds of instances.

PR #1082, opened 2025-02-20
Update go version k8s deps and images

dholbach:


evrardjp (approved):

evrardjp (changes requested):
Seems like the image isn't working in kind CI, do we need to bump kind ?

dholbach:
> Seems like the image isn't working in kind CI, do we need to bump kind ?

Good thinking!

evrardjp (approved):
Test ok

PR #1094, opened 2025-03-03
Migrate virtual instances using virtctl (KubeVirt)

mgrzybek:
This PR adds support for migrating virtual machines managed by KubeVirt.

1. An `Evacuator` interface is used to publish a generic type.
2. A `Kubevirt` evacuator implements this interface.

Other implementations could be added in the future, according to what users need. This could draw a way to solve https://github.com/kubereboot/kured/issues/737 for example.

Concerning KubeVirt SDK usage, two options were tested:

- [ ] Using the native client (https://github.com/kubevirt/client-go/tree/main) broked some Kubernetes library dependancies. Versions conflicts occured when using at the same time the Kubernetes SDK and the KubeVirt one.
- [x] Using `virctl` provided a low dependancy method to interact with KubeVirtks objects. Moreover, the required data could be read directly from Pods' labels.


evrardjp:
I like the principle.
This is fairly close to what I do in another branch, where I refactor the whole main loop.

Would you be okay to wait a bit so we discuss the approach in a next meeting?

mgrzybek:
I think a generic evacuator running shell scripts should be the way to go. The end users would be able to use their on scripts. Moreover this wouldn't claim any official support from any other tool.

About KubeVirt, `evictonStrategy` exists to deal with draining nodes. However this means that some configuration should be made by the users or the k9s admins. I wrote this patch because my users and clusters don't set this value properly. This is a workaround.

I won't maintain this patch on a long-term basis. I can on the other hand write a patch to handle external binaries / shell scripts.

mgrzybek:
Moreover,

Another Evacuator implementation, in addition to a script / binary executor, could be a HTTP caller to trigger some external process.

jeefy:
I like the idea of this being more generic, more akin to the spirit of https://github.com/kubereboot/kured/issues/737

The PR is pretty specific to KubeVirt. I'm curious (and admittedly somewhat ignorant) why there's a need for a custom evacuator in Kured for KubeVirt when there's already a way to specify an eviction strategy to trigger a live migration. Ideally KubeVirt itself should be able to handle a Kubernetes event (drain/eviction) without needing glue in a separate project.

PR #1141, opened 2025-06-05
Fix all references to /var/run to just be /run

cgwalters:
On any modern system `/var/run` should just be a compat symlink to `/run`.

For this project it's probably fine as is, but it can matter in cases where the compat symlink might not be present yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Kured (KUbernetes REboot Daemon) is a Kubernetes daemonset that
 performs safe automatic node reboots when the need to do so is
 indicated by the package management system of the underlying OS.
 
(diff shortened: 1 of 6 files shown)


evrardjp:
Well, this changes the default for many. I am not against it, but i am afraid of breaking expectations. Do you have a use case recently that warrant it? Or should we simply do this to align to modern times?

evrardjp:
We will need to update rebootSentinel in the helm chart too for consistency. 

I would welcome a PR on the helm chart repo containing the change on rebootSentinel and a helm chart version bump.

The helm chart would ideally be merged post release of kured for consistency with cli flags ... but it's optional imo.

PR #1157, opened 2025-07-14
Update k8s for 1.32

evrardjp:
Now that 1.33 kind is released, we can use it to bump to k8s 0.32, which should work for the 3 releases (1.31,1.32,1.33).

This work guarantees it with the help of our CI.


evrardjp:
I will exceptionally self merge due to the absence of review and regular aspect of the change.

evrardjp:
Well well well... it seems GitHub doesn't allow that. 

I will request another review then ...

Issue #1156, opened 2025-07-11
Feature Request: Expose Prometheus metric for current lock holder node


localleon:
Hi team,

I'd like to propose the addition of a new Prometheus metric to expose which node currently holds the reboot lock, for better observability and integration with monitoring systems.

Please consider adding a metric such as:

```bash
kured_current_node_lock{node="node-name"} = 1
```

This metric would:
- Be exposed on all the nodes, regardless if they currently holdthe lock.
- Allow Prometheus/Grafana to surface which node is actively coordinating a reboot.
- Help cluster operators quickly identify reboot activity across nodes.

Why This Is Useful:
-   Debugging coordination issues: Knowing which node is holding the lock helps diagnose stuck or long reboots.
-   Auditing: Helps confirm reboots are progressing as expected across rolling updates.
-   Alerting: We can alert if a lock is held for too long or is stuck on a specific node.

It could be gated behind a feature flag or config option if needed. Since the locking mechanism is already in place via annotations or leases, this metric could easily map to the local node identity.

Happy to discuss more and open to provide a PR if a design decision is made. 
Thanks!

evrardjp:
I don't see why not. Let's add this to v2. However, all those reasons will most likely disappear in v2

evrardjp:
In the demo-ed v2, there is no lock object anymore.  A metric is showing if the node belongs to a maintenance window.  Another if anything was blocked. This can be therefore safely closed when v2 is released.

PR #1164, opened 2025-07-24
Add new kured metrics

colinmcintosh:
Add metrics:
- kured_lock_annotation
- kured_lock_held
- kured_node_draining
- kured_reboot_blocked
- kured_reboot_window_active

This also reduces the interval which metrics are collected from 60s to 15s (4x increase). This reduces the chance of lost metrics during a mismatch of the metrics collection frequency and the metrics scrape frequency.

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -86,10 +86,35 @@ var (
 	annotateNodes bool
 
 	// Metrics
+	lockAnnotationGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: "kured",
+		Name:      "lock_annotation",
+		Help:      "Kured lock annotation metadata.",
+	}, []string{"node", "kured_node_id", "kured_node_unschedulable"})
+	lockHeldGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: "kured",
+		Name:      "lock_held",
+		Help:      "Kured holds a lock allowing a node to reboot.",
+	}, []string{"node"})
+	nodeDrainingGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: "kured",
+		Name:      "node_draining",
+		Help:      "Kured is in the process of draining a node prior to reboot.",
+	}, []string{"node"})
+	rebootBlockedGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: "kured",
+		Name:      "reboot_blocked",
+		Help:      "OS requires reboot but the reboot is blocked by a condition.",
+	}, []string{"node"})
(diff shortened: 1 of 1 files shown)


evrardjp:
Good idea. Let's talk about it in the next community meeting.

localleon:
This would solve some of my concerns in #1156. How does this approach handle conflicting values for the `lock_annotation`? Won't the potentially different update interval of each pod not result in `lock_annotation.node` beeing reported differently by all pods? e.g. Node1 takes the lock -> all pods scrape -> Node 2 takes the lock -> first pod scrapes, but all other pods wait 2 secs for next scrape -> Prometheus metric is incosistent?

localleon:
@evrardjp what would be needed to get this merged?

colinmcintosh:
@localleon thanks for the reminder. Regarding the potential to have inconsistent metrics reported, I did consider that but I think the current implementation provides visibility for kured pods that aren't correctly tracking the lock holder. That should ideally never happen so I'm not convinced that this is the correct implementation. Absolutely open to alternatives if you have strong feelings about it.

@evrardjp happy to discuss in the next community meeting or an adhoc call. We can coordinate in the CNCF slack channel on it if you'd like.

localleon:
@evrardjp hope your doing well! Is there anything I could do so we could get this feature merged?

evrardjp:
I am in the middle of the big rewrite of v2 (you can see the first steps in #1000). 
V2 is also documented in our channel, but I guess I should put a project in github  ... 

Anyway, long story short, most of those metrics won't make sense anymore.
If you're looking into observability, v2 will bring other methods to see what's going on.

Let's go through each metric:
- kured_lock_annotation --> We won't need locks anymore: There will be a shared queue or a lease system between rebooters
- kured_lock_held --> Same comment
- kured_node_draining --> You will be able to observe this through node conditions
- kured_reboot_blocked --> You will be able to observe this through node conditions
- kured_reboot_window_active --> This is something I need to figure out with you. Would you rather have something on the node that says "reboot window is active for this node" (condition/metric/..) or would you prefer that maintenances have their own CRD, which report such status? This was not decided in the v2 document.

Any opinion on this @localleon  @colinmcintosh ?

localleon:
Thanks @evrardjp for the detailled reply! I assumed the PR #1000 was inactive and did not realize that there was a v2 in the works for this project! A Github Project board would probaly be a good idea if your are looking for contributions! 

The metrics look good! About kured_reboot_window_active.. 
- I currently like the simplicity of kured and that it's not using CRDs. I think if possible, i would keep it this way! 
- A metric on each node that says reboot windows is active for this node would be great! From my current understanding we can only configure reboot windows for all nodes at the same time. There are i think two options: 
- - Indicate per node with a 0/1 metric if it's currently able to accept a reboot command 
- - Make a global metric with 0/1 that indicates if all nodes are currently rebootable (this would make more sense since we do not have individual reboot windows)

PR #1215, opened 2025-09-24
Add configuration to block on node labels

stmcginnis:
This proposes adding a new configuration setting to be able to block reboots on nodes that have matching labels applied, similar to the pod selector and annotation locking.

The user can specify one or more `--block-node-label` parameters, using the same kubectl label selector format to match and block on whether a label is present (e.g. `my.custom.label`) or if the label is set to a specific value (e.g. `my.custom.label:block-reboots`).

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -74,6 +74,7 @@ var (
 	messageTemplateReboot           string
 	messageTemplateUncordon         string
 	podSelectors                    []string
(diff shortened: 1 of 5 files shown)


evrardjp:
Would you mind writing about your use case? I would like to have it on the record. This is for history's sake :) 

I think it's also worth doing an e2e test, this way v2 will not break this workflow.

PR #1222, opened 2025-10-03
feat: add multiarch support for hack binaries

dharsanb:
Added multiarch hack binaries to support x86_64 and aarch64 testing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,8 @@ GORELEASER_CMD=$(HACKDIR)/goreleaser
 DH_ORG ?= kubereboot
 VERSION=$(shell git rev-parse --short HEAD)
 SUDO=$(shell docker info >/dev/null 2>&1 || echo "sudo -E")
+ARCH=$(shell uname -m)
+ARCH_COSIGN=$(shell uname -m | sed -e 's/aarch64/arm64/g' -e 's/x86_64/amd64/g')
 
 all: image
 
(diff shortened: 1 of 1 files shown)


evrardjp:
Hello.

Wouldn't this fail on some other binaries, and prevent you from working in aarch64?
What is your development machine?

What do you think about using "mise" or something similar?

dharsanb:
Currently, these are the binaries that are manually downloaded by the Makefile. For other tools like go, I currently download them using package managers so they download the appropriate arch binaries.

My development machine is a M series MacBook with Ubuntu [Lima](https://github.com/lima-vm/lima) VMs for AArch64 and Hetzner VMs for x86-64.
Haven't heard of Mise. But other option is to use Docker images.

Will explore a bit more if we plan on adding more dev-tools.

I think of this PR as a hot fix for development setup for anyone using an ARM machine.

evrardjp:
If it's hotfix, I will merge it as is.

I believe it would be better to use mise, this way we got a simple toml file to update, which can be easily updated using dependabot.

Next to that, mise has a lock file that will help us avoid supply chain issues by refering links and their sha. I believe the tooling will be quite convenient.

I would use mise as a bootstrap tool if the user doesn't have mise already.

WDYT?

dharsanb:
Yes, it is a hotfix. 
I will look into https://github.com/kubereboot/kured/issues/1224 later this week for a comprehensive solution.

Also, don't think Mise is supported by Dependabot yet https://github.com/dependabot/dependabot-core/issues/12320.

dharsanb:
Addressed issues in https://github.com/kubereboot/kured/pull/1227

evrardjp:
Wow you're fast. I intended to merge this , while we take some time for mise. 

Having mise directly is great!

PR #1247, opened 2025-10-31
fix: bump go version to fix vulnerabilities.

dharsanb:
This PR bumps go version to fix security vulnerabilities that causes actions to fail.


evrardjp:
@dharsanb would you mind backporting it to v1?

PR #1262, opened 2025-12-19
Fix trademark usage link to the new redirect

dharsanb:


dharsanb:
Updated Go version to fix vulnerability test failure.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kubereboot/kured
 
-go 1.24.9
+go 1.24.11
```
evrardjp: Could you remove this from this commit please?

evrardjp:
Else I agree, and thank you a lot for this :)  Good catch.

go.mod
dharsanb: Action run [fails](https://github.com/kubereboot/kured/actions/runs/20356168212/job/58492051178) as there's a vulnerability in this go version.
Should I commit in a different PR?

PR #1264, opened 2025-12-20
Modify Dockerfile to use artifacts.json instead of custom logic

dharsanb:
This is done to support v2 and future releases of Goreleaser.


evrardjp:
I am worried this might break the multi container build that we'll do in v2? How can I test this?

dharsanb:
I tested it using 
```shell
 goreleaser release --snapshot --clean
 docker buildx build --platform linux/arm64,linux/amd64,linux/arm/v7,linux/arm/v6,linux/386 -t kured --no-cache .
```
This creates binaries using goreleaser under dist and I trigger docker builds manually for multiple platforms to make sure the builds work as expected .

evrardjp:
@dharsanb do you want me to merge this? I like it. I just wonder whether goreleaser can be updated further more as they introduced the new feature you requested?

dharsanb:
Think this should work. The new feature did not work as I expected. I think it only creates those files when you do the actual build.

evrardjp:
We should discuss this - i am not sure I got your point :( 

currently the dev build is mono arch indeed, amd64, but it can be improved ?

dharsanb:
This PR is only applicable for Docker builds using goreleaser binaries. So we can close this issue.

PR #1267, opened 2026-01-06
Update Kured to support kubernetes 1.35

evrardjp:
Now that a new kind version is released supporting 1.35, we can use it for our CI testing.

This commit:
- Bumps the kind version and its images to support the 1.35
- Ensure all the API calls are done with client-go (and other k8s deps) supporting a maximum ranges of versions (0.34)
- Aligns Kured go version with k8s 0.34 go version.


evrardjp:
This will update on main branch (for quick releasing), then we can backport on v1 branch for long term support

evrardjp:
Rebased.

PR #1268, opened 2026-01-06
Update policy on kube versions

evrardjp:
For the future! (This needs to be rebased on top of main when #1267 is merged)


evrardjp:
Need rebase.

evrardjp:
We need to merge this as is and fix the actions not bumping trivy. 

The latter will fail if this one hasn't merged ...

PR #1344, opened 2026-06-08
Update Kured to latest kubernetes modules

evrardjp:
Now that a new kind version is released supporting 1.36, we can use it for our CI testing.

This commit:
- Bumps the kind version and its images to support the 1.36
- Ensure all the API calls are done with client-go 0.36 (and dependent k8s deps), to allow future 1.37 release.


evrardjp:
Requires #1347 to pass Short tests.

evrardjp:
@dharsanb can you have a look at this? I merged, but I would love to get your approval, even after the fact.

PR #1348, opened 2026-06-08
Bump goreleaser to v2 (alt)

evrardjp:


evrardjp:
The "fix: Stop building main branch containers" should help on the following issue:

```
WARNING: Image reference ghcr.io/kubereboot/kured:20b968b uses a tag, not a digest, to identify the image to sign.
    This can lead you to sign a different image than the intended one. Please use a
    digest (example.com/ubuntu@sha256:abc123...) rather than tag
    (example.com/ubuntu:latest) for the input to cosign. The ability to refer to
    images by tag will be removed in a future release.
```

I hope by moving to go releaser fully only on release, this will go away too.

dharsanb (approved):
LGTM

PR #1355, opened 2026-06-09
Simplify testing

evrardjp:


evrardjp:
@dharsanb this is a follow up to my previous commits -- It will clean the tests further. Less code to maintain!

PR #1361, opened 2026-06-11
Remove the redundant DH_ORG environment variable 

dharsanb:
In this PR:
1. DH_ORG environment variable has been removed as it's only used in the makefile to generate image name
2. Added registry as a configurable variable in case someone is using a registry other than GitHub.

BREAKING-CHANGE: This breaks the existing development environment where DH_ORG and IMAGE_NAME was configured. This should be changed in development machines. Can document the same in the developer documentation.


evrardjp (approved):
I like it

PR #1372, opened 2026-06-21
validate empty sentinel file path in NewFileRebootChecker

Ahmad-Faraj:
NewFileRebootChecker already returns an error but never actually used it. if you pass an empty --reboot-sentinel the checker just stats "" forever and silently never reports a reboot, which is annoying to debug.

so this rejects an empty/blank path up front instead. dropped the old TODO and added a small test.


evrardjp (approved):

evrardjp (changes requested):
let's follow copilot suggestion

Ahmad-Faraj:
Done, trimming once and storing the trimmed path now.

Ahmad-Faraj:
@evrardjp the trim change is in, `157fcaeb`. Any chance of another look?

evrardjp:
I have to put that into the context of the other changes, but yeah I will work on it when I can. Thanks for your contribution.

PR #1388, opened 2026-07-04
fix: reject node labels not in key=value form to prevent uncordon panic

arpitjain099:
This fixes the crash in #1281. If a pre- or post-reboot node label is passed without an `=` (the reporter used a bare `node.kubernetes.io/exclude-from-external-load-balancers-`), `updateNodeLabels` does `strings.Split(label, "=")[1]` on a single-element slice and panics with an index out of range, taking kured down right after the node is uncordoned.

I moved the label parsing to `SplitN` and made `validateNodeLabels` reject anything that isn't in `key=value` form, so a bad value is caught at startup instead of mid-reboot where it can strand a node. `updateNodeLabels` now skips a malformed entry rather than panicking as a second line of defense. Empty values like `key=` are still accepted since Kubernetes allows empty label values. I kept this to the panic itself; the label-removal syntax the reporter was reaching for (a trailing `-`) is a separate feature and out of scope here.

Added a `TestValidateNodeLabels` table covering the no-`=` and empty-key cases; it fails on main today and passes with the fix, and `go test ./cmd/kured/...` is green. Thanks for taking a look.

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -6,6 +6,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"math/rand"
 	"net/http"
@@ -222,6 +223,12 @@ func main() {
 
 	err := validateNodeLabels(preRebootNodeLabels, postRebootNodeLabels)
 	if err != nil {
+		// A malformed label (not in key=value form) would panic later during
+		// uncordon, so refuse to start. A mere key mismatch between the pre and
+		// post reboot sets is only surprising, not fatal, so keep warning.
+		if errors.Is(err, errMalformedNodeLabel) {
+			log.Fatal(err.Error())
+		}
 		log.Warn(err.Error())
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/kured/main.go
evrardjp, line 232: why not fataling here then? Isn't it better than silently ignoring something the user has set? I am worried about expectations. 

At the same time, it's gonna break some people. But breaking some people is probably what we want if they got something misconfigured. 

So a good release note + fataling seems fine for me.

WDYT @dharsanb @jackfrancis ?


```suggestion
		log.Fatal(err)
```

evrardjp: Alternatively, do an error check: 
If it's just medium severity mistmatch of node labels, continue to warn, as usual, but if it's MALFORMED nodelabel, just fatal. 

so we might want something like
```
var (
    errMalformedNodeLabel = errors.New("malformed node label")
    errMismatchedNodeLabelKeys = errors.New("pre/post reboot node label keys do not match")
)
```

So in splitNodeLabel:
```
fmt.Errorf("%w: %q must be in key=value form", errMalformedNodeLabel, label)
```

 In validateNodeLabels,  wrap the mismatched/malformed errors...

Then main becomes

```
err := validateNodeLabels(preRebootNodeLabels, postRebootNodeLabels)
if err != nil { 
  if errors.Is(err, errMalformedNodeLabel) { 
    log.Fatal(err.Error())
  }
  log.Warn(err.Error())
}       
```

arpitjain099:
Thanks @evrardjp, that split makes a lot of sense. I've pushed a commit that adds the two typed sentinels you sketched, `errMalformedNodeLabel` and `errMismatchedNodeLabelKeys`. `splitNodeLabel` now wraps the malformed one with `fmt.Errorf("%w: %q must be in key=value form", ...)` and the key-mismatch path wraps the other, so `main` tells them apart with `errors.Is`: a malformed label is `log.Fatal` (it's the one that panics later during uncordon, better to refuse to start), and a plain key mismatch stays a `log.Warn` like before.

I kept the tests asserting the wrapped sentinel per case via `errors.Is` so the fatal-vs-warn split is pinned down, and `go test ./cmd/kured/...` plus `go vet` are green. Happy to add a release note about the malformed-label case now being fatal if you'd like it in this PR. Let me know if you'd rather fatal on the mismatch too once @dharsanb / @jackfrancis weigh in.

PR #1395, opened 2026-07-28
Add cooldowns to dependabots

evrardjp:
To avoid wrong bumps that are quickly fixed, or supply chain attacks, we might want to have a grace period in all our dependabot updates.

This introduces it.


dharsanb:
Should we add it to docker also?

evrardjp:
Yup. I will do it.

PR #1396, opened 2026-07-28
fix CVE-2026-39822

evrardjp:
Without this, trivy will fail.


evrardjp:
It failed to queue again, not sure why... I stopped all the other PRs in queue though... Let me retry the whole thing

PR #1397, opened 2026-07-28
Introduce AGENT guidance

evrardjp:
Without this, people will do many things in our code base that are not appropriate.

This lets us have a framework.


evrardjp:
This needs #1000 to be in first.

PR #1409, opened 2026-08-14
Uncordon node before releasing lock after a failed drain

### LINKED ISSUE #1129: Uncordon immediately after drain failure
We've been running Kured for some time and recently noticed a change in behavior starting with version 1.16.2. When a node fails to drain, it is now uncordoned only after the releaseDelay period.

Previously (e.g., in version 1.16.1 and earlier), the node was uncordoned immediately after the drain failure. This behavior was preferable for our use case, as it allows workloads to be rescheduled quickly if a reboot cannot proceed.

It appears this change was introduced in [this commit](https://github.com/kubereboot/kured/commit/104a745305e02d85ce1555b7d2c7d1f44cab1ccf).

We would like to request an option or a fix to restore the previous behavior—immediate uncordon on drain failure—to minimize disruption in scenarios where the reboot cannot be completed.

pujitha24:
Motivation:
When a node fails to drain and forceReboot is disabled, kured releases
the lock and then performs a best-effort uncordon so the node can keep
receiving workloads. Both DaemonSetSingleLock.Release() and
DaemonSetMultiLock.Release() (pkg/daemonsetlock/daemonsetlock.go) sleep
for the configured --lock-release-delay before doing anything else.
Because uncordon was called only after lock.Release() returned, the
node stayed cordoned for the entire release delay even though no
reboot was going to happen, defeating the purpose of a best-effort
uncordon on drain failure. This regressed behavior present before
commit 104a745305e02d85ce1555b7d2c7d1f44cab1ccf, which moved the
release delay into Release() without reordering the caller.

Approach:
Extract the drain-failure handling in rebootAsRequired into a new
uncordonThenRelease helper (cmd/kured/main.go) that uncordons the node
first and only then releases the lock, so the release delay no longer
blocks the uncordon. The lock is still released even if uncordon
fails, matching the previous error-handling behaviour.

Validation:
- go build ./...
- make test (lint-sh, lint-ghactions, golangci-lint, and
  `go test -test.short ./...`) passes with 0 lint issues and 0 test
  failures.
- Added TestUncordonThenReleaseUncordonsBeforeReleasingLock and
  TestUncordonThenReleaseReleasesLockEvenIfUncordonFails
  (cmd/kured/main_test.go), using a minimal fake daemonsetlock.Lock to
  assert call order. Confirmed these are a real regression test: with
  the order in uncordonThenRelease temporarily reverted to
  release-then-uncordon, TestUncordonThenReleaseUncordonsBeforeReleasingLock
  fails ("call order = [release uncordon], want [uncordon release]");
  restoring the fix makes it pass again.
- Not run: the kind-based e2e suite (make e2e-test), which needs a
  local kind/docker cluster not available in this environment.
  CONTRIBUTING.md describes e2e tests as encouraged for functional
  testing, not required for a fix of this size, and `make test` is the
  documented CI-mirroring command.

Report: https://github.com/kubereboot/kured/issues/1129
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

Fixes #1129

### DIFF
--- a/cmd/kured/main.go
+++ b/cmd/kured/main.go
@@ -479,6 +479,19 @@ func uncordon(client *kubernetes.Clientset, node *v1.Node) error {
 	return nil
 }
 
+// uncordonThenRelease performs a best-effort uncordon before releasing the lock, so that a
+// failed drain does not leave the node cordoned for the duration of any configured lock
+// release delay.
+func uncordonThenRelease(nodeName string, lock daemonsetlock.Lock, uncordonFn func() error) {
+	log.Infof("Performing a best-effort uncordon after failed cordon and drain")
+	if err := uncordonFn(); err != nil {
+		log.Errorf("Failed to uncordon %s: %v", nodeName, err)
+	}
+	if err := lock.Release(); err != nil {
+		log.Errorf("Error releasing lock: %v", err)
+	}
+}
+
 func maintainRebootRequiredMetric(nodeID string, checker checkers.Checker) {
 	for {
 		if checker.RebootRequired() {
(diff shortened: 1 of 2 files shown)


pujitha24:
This is rebased and green whenever someone has a chance to take a look — happy to make any changes that would help review go smoother.

pujitha24:
Closing this to keep my open PR queue manageable - it has been open a while without review, and I would rather not leave stale PRs sitting in your queue. The change itself still applies; happy to reopen and rebase if it is useful to you.

